PR #7, opened 2024-06-21
Update to numpy 2.0

adityagoel4512:
With the release of NumPy 2.0 earlier this week, [Array API support was promoted](https://numpy.org/doc/stable/release/2.0.0-notes.html#highlights) to the main ``numpy`` namespace.

This updates the lockfile. It also updates the documentation and tests to remove the use of ``numpy.array_api``.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,7 @@ jobs:
           - py310
           - py311
           - py312
+          - np1x
     steps:
       - name: Checkout branch
         uses: actions/checkout@v4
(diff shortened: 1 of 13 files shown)


cbourjau:
Would it make sense to test on `numpy<2` for a few more months?

adityagoel4512:
> Would it make sense to test on numpy<2 for a few more months?

Definitely, I added an environment for NumPy 1.x.

cbourjau (approved):
Pixi ❤️ !

PR #14, opened 2024-07-10
Fix public export promote_nullable

EmilyMohellenbrandQC:
## Bug
`promote_nullable` is publicly exposed, but it's not working because it is not imported properly in the init file.

## Changes
Add import `promote_nullable` in `ndonnx.__init__`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,13 @@
 Changelog
 =========
 
+0.5.1 (unreleased)
+------------------
+
+**Bug fixes**
+
+- Public export of :func:`ndonnx.promote_nullable` is now working.
+
 0.5.0 (2024-07-01)
 ------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,13 +6,12 @@
 Changelog
 =========
 
-0.5.1 (unreleased)
+0.6.0 (unreleased)
 ------------------
 
-**Bug fixes**
+**Other changes**
 
-- Public export of :func:`ndonnx.promote_nullable` is now working.
-
+- :func:`ndonnx.promote_nullable` is now publicly exported
 0.5.0 (2024-07-01)
 ------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@
 Changelog
 =========
 
-0.6.0 (unreleased)
+0.6.0 (2024-07-11)
 ------------------
 
 **Other changes**
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ Changelog
 **Other changes**
 
 - :func:`ndonnx.promote_nullable` is now publicly exported
+
 0.5.0 (2024-07-01)
 ------------------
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -6,6 +6,13 @@
 Changelog
 =========
 
+0.5.1 (unreleased)
+------------------
+
+**Bug fixes**
+
+- Public export of :func:`ndonnx.promote_nullable` is now working.
+
```
adityagoel4512, line 15:
```suggestion
0.6.0 (unreleased)
------------------

**Other changes**

- :func:`ndonnx.promote_nullable` is now publicly exported
```

I don't think this is a bug given it is explictly used in the tests through private APIs even in the tests. This would simply make it public.

adityagoel4512: Side note - before doing this, I would like to enable user-defined data types to define their "nullable" counterpart.

cbourjau: @adityagoel4512 One way or another there was a bug here: We were having `promote_nullable` in `__all__` but didn't import it into the scope.

cbourjau (dismissed):
I'm not particularly happy with this function; being explicit about the mask as [`make_nullable`](https://github.com/Quantco/ndonnx/blob/9eb981ad801f6f5294958ea0ea6a5ad28879fac4/ndonnx/additional/_additional.py#L120) is doing seems better. For the sake of stability, we may make an exercise out of this though: Let's fix the export but deprecate it immediately in favor of `make_nullable`. Then we remove it in a release or two from the public interface.

CHANGELOG.rst
EmilyMohellenbrandQC: @adityagoel4512 judging by the `__init__` file alone I would think this is a bug, because the function is already listed in `__all__` but only the import was not working. But I'm ok with either way how we call this change.

adityagoel4512: I never disagreed that the import did not work. Adding to the public API (which this does) is by definition a minor change however.

CHANGELOG.rst
```diff
@@ -6,6 +6,12 @@
 Changelog
 =========
 
+0.6.0 (unreleased)
+------------------
+
+**Other changes**
+
+- :func:`ndonnx.promote_nullable` is now publicly exported
```
EmilyMohellenbrandQC, line 14:
```suggestion
- :func:`ndonnx.promote_nullable` is now publicly exported

```

adityagoel4512:
I agree! Can you please add a DeprecationWarning @EmilyMohellenbrandQC?

ndonnx/_data_types/__init__.py
```diff
@@ -69,6 +71,12 @@ def promote_nullable(dtype: StructType | CoreType) -> _NullableCore:
         If the input type is unknown to ``ndonnx``.
     """
 
+    warn(
+        "Function 'ndonnx.promote_nullable' is to be deprecated. "
```
adityagoel4512, line 75:
```suggestion
        "Function 'ndonnx.promote_nullable' will be deprecated in ndonnx 0.7. "
```

adityagoel4512 (approved):

cbourjau:
When deprecating a function we should make sure that we don't use it ourselves in this code base, too. Otherwise, we thrust deprecation warnings into the users face that they can do nothing about. Unfortunately, this happened here: https://github.com/Quantco/ndonnx/actions/runs/9892611298/job/27325701365#step:5:41

PR #17, opened 2024-07-16
Allow user defined data type operations with core types

adityagoel4512:
Currently, if you define your own data type as a user (an experimental feature!), you can happily implement many functions especially if the operands are themselves user-defined data types. However, if you do a binary operation like addition between a i64 and a custom type, type promotion fails since arbitrary dtypes cannot safely be added to the [type-promotion hierarchy](https://data-apis.org/array-api/latest/API_specification/type_promotion.html#type-promotion). Promotion occurs in the implementation of `add` and many other functions for the types exposed by ndonnx.

We should retain this restriction on type promotion but allow custom type implementers to implement binary operations that interact directly with Array API dtypes.

cc @MatejUrbanQC

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,8 +11,7 @@ Changelog
 
 **New features**
 
-- ``result_type`` now raises a ``PromotionError`` if the input types are not compatible.
-
+- :func:`~ndonnx.result_type` now raises a :class:`~ndonnx.PromotionError` if the input types are not compatible.
 
 0.6.1 (2024-07-12)
 ------------------
--- a/ndonnx/_funcs.py
+++ b/ndonnx/_funcs.py
@@ -290,7 +290,7 @@ def result_type(
                 np_dtypes.append(dtype.values.to_numpy_dtype())
             else:
                 raise PromotionError(
-                    "result_type is not defined for arbitrary Struct types. You likely need to explicitly cast to the desired dtype."
+                    "'result_type' is not defined for arbitrary struct types. You likely need to cast to the desired dtype explicitly."
                 )
         else:
             np_dtypes.append(dtype.to_numpy_dtype())
--- a/ndonnx/_funcs.py
+++ b/ndonnx/_funcs.py
@@ -20,6 +20,7 @@ from ndonnx._data_types.structtype import StructType
 from . import _opset_extensions as opx
 from ._array import Array, _from_corearray
 from ._utility import (
+    PromotionError,
     promote,
 )
 
@@ -279,8 +280,11 @@ def isdtype(dtype, kind):
 def result_type(
     *objects: dtypes.CoreType | dtypes.StructType | Array,
 ) -> dtypes.CoreType | dtypes.StructType:
-    """Find the common data type for the provided objects."""
     observed_dtypes = {obj.dtype if isinstance(obj, Array) else obj for obj in objects}
+
+    if len(observed_dtypes) == 1:
+        return next(iter(observed_dtypes))
+
     nullable = False
     np_dtypes = []
     for dtype in observed_dtypes:
@@ -289,7 +293,9 @@ def result_type(
                 nullable = True
                 np_dtypes.append(dtype.values.to_numpy_dtype())
             else:
-                raise TypeError("result_type is not defined for arbitrary Struct types")
+                raise PromotionError(
+                    "result_type is not defined for arbitrary Struct types. You likely need to explicitly cast to the desired dtype."
+                )
         else:
             np_dtypes.append(dtype.to_numpy_dtype())
 
@@ -386,10 +392,17 @@ def equal(x, y):
 def _binary(func_name, x, y):
     x_dtype = asarray(x).dtype
     y_dtype = asarray(y).dtype
-    if (out := getattr(x_dtype._ops, func_name)(x, y)) is not NotImplemented:
-        return out
-    if (out := getattr(y_dtype._ops, func_name)(x, y)) is not NotImplemented:
-        return out
+    try:
+        if (out := getattr(x_dtype._ops, func_name)(x, y)) is not NotImplemented:
+            return out
+    except PromotionError:
+        pass
+
+    try:
+        if (out := getattr(y_dtype._ops, func_name)(x, y)) is not NotImplemented:
+            return out
+    except PromotionError:
+        pass
     raise TypeError(
         f"Unsupported operand type(s) for {func_name}: '{x_dtype}' and '{y_dtype}'"
     )
--- a/ndonnx/_funcs.py
+++ b/ndonnx/_funcs.py
@@ -281,10 +281,6 @@ def result_type(
     *objects: dtypes.CoreType | dtypes.StructType | Array,
 ) -> dtypes.CoreType | dtypes.StructType:
     observed_dtypes = {obj.dtype if isinstance(obj, Array) else obj for obj in objects}
-
-    if len(observed_dtypes) == 1:
-        return next(iter(observed_dtypes))
-
     nullable = False
     np_dtypes = []
     for dtype in observed_dtypes:
--- a/ndonnx/_funcs.py
+++ b/ndonnx/_funcs.py
@@ -410,8 +410,11 @@ def _binary(func_name, x, y):
 
 def _unary(func_name, x):
     x = asarray(x)
-    if (out := getattr(x.dtype._ops, func_name)(x)) is not NotImplemented:
-        return out
+    try:
+        if (out := getattr(x.dtype._ops, func_name)(x)) is not NotImplemented:
+            return out
+    except PromotionError:
+        pass
     raise TypeError(f"Unsupported operand type for {func_name}: '{x.dtype}'")
 
 
--- a/ndonnx/_utility.py
+++ b/ndonnx/_utility.py
@@ -9,15 +9,20 @@ import numpy as np
 import numpy.typing as npt
 from spox import Var
 
-# FIXME: Remove private import from Spox! Better to just use reshape!
-from spox._internal_op import unsafe_reshape
-
+import ndonnx as ndx
 import ndonnx._data_types as dtypes
 
 if TYPE_CHECKING:
     from ndonnx import Array
 
     from ._corearray import _CoreArray
+
+
+class PromotionError(TypeError):
+    """Error raised when type promotion fails."""
+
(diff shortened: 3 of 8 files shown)


adityagoel4512:
@MatejUrbanQC

ndonnx/_utility.py
```diff
@@ -64,17 +58,22 @@ def _promote_with_none(*args: Array | npt.ArrayLike) -> list[Array | None]:
...
+
+    if target_dtype is None:
+        return None
 
-    return [None if arr is None else astype(arr, target_dtype) for arr in arr_or_none]
+    return [None if arr is None else arr.astype(target_dtype) for arr in arr_or_none]
 
 
-def promote(*args: Array | npt.ArrayLike) -> list[Array]:
+def promote(*args: Array | npt.ArrayLike) -> list[Array] | None:
```
cbourjau, line 69: This signature makes it very annoying to work with from a typing point of view. Not having `None` in the return type is the exact reason why we have `promote` and `_promote_with_none`.

cbourjau (commented):
This PR would be much easier to review if it weren't blended with various unrelated clean-up chores. 

Could you maybe elaborate on the use case in which this came up, please?

ndonnx/_utility.py
adityagoel4512: I think this is an aesthetic decision. The other options are an explicit result class or driving control flow through exception handling. I'm not sure either of those solutions is super pleasant.

adityagoel4512:
> This PR would be much easier to review if it weren't blended with various unrelated clean-up chores.

The other thing that changed is that in `_promote_with_none` the lazy importing was removed for top-level import and an unused function removed. If it helps to move that out of this PR that's also fine but it felt local to the changes being made anyway.

> Could you maybe elaborate on the use case in which this came up, please?
This occurred when adding a custom data type with a core data type (specifically in @MatejUrbanQC's exploratory work).

ndonnx/_utility.py
cbourjau: But it will nag us for every promotion between built-in data types. `None` is not very descriptive either. I think the most Pythonic thing would indeed be an exception.

cbourjau:
A while ago we set out to move the logic for the individual operations into the data types. One may argue that promotion is a core part of that logic, yet we left it in the generic top-level functions. Would it make sense for this feature to move the promotion into the data type member functions? Is there a reason we didn't do that last time around? 

 I also think the code would be cleaner if we would use an exception for a failed promotion rather than trying to propagate `None` from every propagation call.
 
 > The other thing that changed is that in _promote_with_none the lazy importing was removed for top-level import and an unused function removed. If it helps to move that out of this PR that's also fine but it seemed very local to the changes being made anyway.
 
 This PR proposes a far-reaching and fundamental change, yet a large part of the diff is unrelated or somewhat gratuitous (e.g. moving to `operants` and `_variadic_op`). It is not ideal.

ndonnx/_utility.py
adityagoel4512: > I think the most Pythonic thing would indeed be an exception.

This PR was originally structured with this. However, it requires exposing a new exception type (there was a `PromotionError` since it is not safe to drive control-flow from an arbitrary `TypeError` down the stack. It also makes the logic in [`_binary`](https://github.com/Quantco/ndonnx/pull/17/files#diff-cd5d4a98029b4c5c571094c69a158fae5d5ee688a5e085614c13f68230984555R386) quite hacky with try-catches wrapping the function calls.

> But it will nag us for every promotion between built-in data types.

I agree that the ergonomics are unfortunate and I was hoping for something neater but I would really like to not treat built-in dtypes as anything special. You should be handling failure anyway if you're using `result_type` if the intention is to write safe code - type-promotion does not need to succeed always. Happy to think about it some more.

adityagoel4512:
> Would it make sense for this feature to move the promotion into the data type member functions? Is there a reason we didn't do that last time around?

I haven't really thought about how user-defined data types would safely integrate themselves into a promotion hierarchy. It's not trivial since you can easily create looping dependencies. We left it out for this reason.

> This PR proposes a far-reaching and fundamental change, yet a large part of the diff is unrelated or somewhat gratuitous (e.g. moving to operants and _variadic_op).

The operands is not unrelated? You cannot destructure the output of `promote` immediately as before if the output could be `None`.

I'm happy to revert back to exceptions.

ndonnx/_utility.py
cbourjau: Indeed, a custom exception would probably be good to have here. Maybe it is ok to expose that exception publicly. NumPy has [`DTypePromotionError`](https://numpy.org/doc/stable/reference/generated/numpy.exceptions.DTypePromotionError.html#numpy.exceptions.DTypePromotionError) in its public interface, too.

ndonnx/_funcs.py
```diff
@@ -279,7 +280,6 @@ def isdtype(dtype, kind):
 def result_type(
     *objects: dtypes.CoreType | dtypes.StructType | Array,
 ) -> dtypes.CoreType | dtypes.StructType:
-    """Find the common data type for the provided objects."""
```
cbourjau, line 282: Was that deliberate?

CHANGELOG.rst
```diff
@@ -6,6 +6,14 @@
 Changelog
 =========
 
+0.7.0 (unreleased)
+------------------
+
+**New features**
+
+- ``result_type`` now raises a ``PromotionError`` if the input types are not compatible.
```
cbourjau, line 14:
```suggestion
- :func:`~ndonnx.result_type` now raises a :class:`~ndonnx.PromotionError` if the input types are not compatible.
```

ndonnx/_funcs.py
```diff
@@ -289,7 +289,9 @@ def result_type(
                 nullable = True
                 np_dtypes.append(dtype.values.to_numpy_dtype())
             else:
-                raise TypeError("result_type is not defined for arbitrary Struct types")
+                raise PromotionError(
+                    "result_type is not defined for arbitrary Struct types. You likely need to explicitly cast to the desired dtype."
```
cbourjau, line 293:
```suggestion
                    "'result_type' is not defined for arbitrary struct types. You likely need to cast to the desired data type explicitly."
```

ndonnx/_funcs.py
```diff
@@ -386,19 +388,29 @@ def equal(x, y):
...
+            return out
+    except PromotionError:
+        pass
+
+    try:
+        if (out := getattr(y_dtype._ops, func_name)(x, y)) is not NotImplemented:
+            return out
+    except PromotionError:
+        pass
     raise TypeError(
```
cbourjau, line 402: Would it be reasonable to simply pass the `PromotionError` through?

tests/ndonnx/test_dtypes.py
```diff
@@ -22,28 +22,51 @@ class Unsigned96Impl(OperationsBlock):
...
+            return y + x
+        elif x.dtype == Unsigned96() and y.dtype == ndx.uint32:
+            return x + y.astype(Unsigned96())
+        return NotImplemented
+
 
 class Unsigned96(StructType, CastMixin):
     def _fields(self) -> dict[str, StructType | CoreType]:
         return {
-            "upper": ndx.uint32,
```
cbourjau, line 29: Is there some deeper context that I'm missing that made this change necessary or why is one ordering preferable to the other?

ndonnx/_funcs.py
adityagoel4512: Yes, it is overwritten https://github.com/Quantco/ndonnx/blob/main/ndonnx/_funcs.py#L1038-L1041.

ndonnx/_funcs.py
adityagoel4512: I don't think so. You would end up getting different errors if you do `a + b` vs `b + a` where one is a core data type and where one is user-defined. Since we removed the explicit management of failure from the functions themselves and use the `PromotionError` to drive dispatching, you cannot raise on the first one so you shouldn't on the second either.

tests/ndonnx/test_dtypes.py
adityagoel4512: No deep meaning - it just makes it possible to get the carry without overflowing for the addition implementation for the tests here.

ndonnx/_funcs.py
cbourjau: But why are we not re-raising the caught `PromotionError`?

cbourjau (commented):
I'm still trying to wrap my head around how this works. Let me put it in my own words:
- Built-in data types call `promote` for function such as `add`, but custom data types bring their own `add` implementation
- The public `ndonnx.add` function calls `_binary` which attempts to call the dtype's `add` function
- If the first operand's `add` fails with `PromotionError`, it tries the second one

Observations:
- We can't unconditionally call `promote` in `ndonnx.add` since if and how we promote may depend on the operation. Sometimes the operands must not have the same type (e.g. adding a time delta to a date time).
- `promote` and `result_type` are only involved in so far that they throw an error if custom data types are involved
- Custom types are not in the regular type hierarchy and we don't plan to allow that at this point
- Type compatibilities depend on the operation and thus must be resolved somehow inside the operations of the custom types
- Built-in types don't ever know how to deal with custom data types. Can we not catch them earlier than when hitting `result_type`?

Might the following work:
- `ndonnx._funcs._binary` never passes a custom data type into a built-in op. Alternatively, all core data type ops may gain a decorator that checks for custom types and raises `NotImplemented` or something.
- We only call custom data type ops with custom data types and they may do as they please.

ndonnx/_funcs.py
adityagoel4512: I think it's best we speak in person/call.

adityagoel4512: The fact that `promote` is being called is an implementation detail of the operation. A type can implement a binary operator with or without type promotion - this is also why I'm not a huge fan of driving this dispatching through exceptions, as I've mentioned elsewhere.

I think it's best we speak directly.

adityagoel4512:
I wouldn't put too much into decorators but some validation does seem appropriate. We should also make pulling in default implementations for "transmute" operations cleaner.

I'll rework things later in the week.

cbourjau:
> I wouldn't put too much into decorators but some validation does seem appropriate.

I 100% agree! Historically, we did A LOT in decorators in this project and we removed it for a reason. I generally think that mutating arguments in decorators should be avoided.

PR #19, opened 2024-07-17
Fix broadcast_arrays for user-defined data types

MatejUrbanQC:
zeros_like needs an explicite dtype, otherwise it will try to cast the value to the dtype of the original array. This can fail for user-defined data types.


adityagoel4512 (approved):
Thanks. I cannot think of anything more concise off the top of my head but it is a bit of a hack to determine the target shape by using the add operator's bidirectional broadcasting - worth exploring better options in the future.

PR #22, opened 2024-07-22
Cast back to input dtype in `where`

adityagoel4512:
Previously `ndx.where(ndx.asarray([True, False]), ndx.asarray([0, 1], ndx.int8), ndx.asarray([2, 3], ndx.int8))` would yield an `int32` output. This upcasting is due to the fact that onnxruntime does not implement `int8` and `int16` kernels, but a respective downcast should exist so the output of `ndx.where` has the dtype of its promoted arguments.


cbourjau (approved):
Should we add a test?

PR #23, opened 2024-07-22
Use ``float32`` in ``where``

MatejUrbanQC:
Onnxruntime doesn't implement ``ReduceProd`` for ``float64``. We will error for
``float64`` and otherwise use ``float32``.

Also fixes the implementation for unsigned integers

### DIFF
--- a/ndonnx/_core/_impl.py
+++ b/ndonnx/_core/_impl.py
@@ -521,6 +521,14 @@ class CoreOperationsImpl(OperationsBlock):
                         b.astype(dtypes.int32)._core(),
                     )
                 )
+            elif a.dtype in (dtypes.uint16, dtypes.uint32, dtypes.uint64):
+                return _from_corearray(
+                    opx.where(
+                        condition_values._core(),
+                        a.astype(dtypes.int64)._core(),
+                        b.astype(dtypes.int64)._core(),
+                    )
+                ).astype(a.dtype)
             elif isinstance(a.dtype, dtypes.CoreType):
                 return _from_corearray(
                     opx.where(
--- a/ndonnx/_core/_impl.py
+++ b/ndonnx/_core/_impl.py
@@ -1051,7 +1051,7 @@ def _via_upcast(
             via_dtype = int_dtype
         else:
             raise TypeError(
-                f"Can't upcast unsigned type {dtype}. Available implementations are for {*available_types,}"
+                f"Can't upcast unsigned type `{dtype}`. Available implementations are for `{*available_types,}`"
             )
     elif isinstance(dtype, (dtypes.Integral, dtypes.NullableIntegral)) or dtype in (
         dtypes.nbool,
--- a/ndonnx/_core/_impl.py
+++ b/ndonnx/_core/_impl.py
@@ -733,7 +733,7 @@ class CoreOperationsImpl(OperationsBlock):
         if dtype is not None:
             x = x.astype(dtype)
 
-        out = _via_i64_f32(
+        out = _via_upcast(
             lambda x: opx.reduce_prod(
                 x,
                 opx.const(axes, dtype=dtypes.int64),
@@ -741,6 +741,8 @@ class CoreOperationsImpl(OperationsBlock):
                 noop_with_empty_axes=axis is not None,
             ),
             [x],
+            int_dtype=dtypes.int64,
+            float_dtype=dtypes.float32,
         )
 
         if isinstance(x.dtype, dtypes.Unsigned):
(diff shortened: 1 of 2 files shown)


ndonnx/_core/_impl.py
```diff
@@ -1029,6 +1037,34 @@ def _via_i64_f64(
     return _variadic_op(arrays, fn, via_dtype, cast_return)
 
 
+def _via_i64_f32(
```
adityagoel4512, line 1040: Can you share some of the logic between this and `_via_i64_f64`. Specifically, I think providing the types as input and making sure you don't downcast (you can check `ndx.iinfo(dtype).bits`/`ndx.finfo(dtype).bits`) is sufficiently general. It would then make it easier to do more upcasting permutations as and when we find the need.

MatejUrbanQC: Yeah makes sense, I'll do that.

ndonnx/_core/_impl.py
```diff
@@ -521,6 +521,14 @@ def where_dtype_agnostic(a: ndx.Array, b: ndx.Array) -> ndx.Array:
                         b.astype(dtypes.int32)._core(),
                     )
                 )
+            elif a.dtype in (dtypes.uint16, dtypes.uint32, dtypes.uint64):
```
adityagoel4512, line 524:
```suggestion
            elif a.dtype in (dtypes.uint16, dtypes.uint32):
```

MatejUrbanQC: I think in this case it's actually safe. There is no loss, the onnx standard defines cast between signed and unsigned as just reinterpreting the bits. It is only unsafe for arithmetic

ndonnx/_core/_impl.py
```diff
@@ -1003,32 +1013,90 @@ def _via_dtype(
...
+            via_dtype = uint_dtype
+        elif int_dtype is not None and ndx.iinfo(dtype).bits <= (
+            ndx.iinfo(int_dtype).bits
+            if use_unsafe_uint_cast
+            else ndx.iinfo(int_dtype).bits - 1
+        ):
+            via_dtype = int_dtype
+        else:
+            raise TypeError(
+                f"Can't upcast unsigned type {dtype}. Available implementations are for {*available_types,}"
```
cbourjau, line 1054:
```suggestion
                f"Can't upcast unsigned type `{dtype}`. Available implementations are for `{*available_types,}`"
```

ndonnx/_core/_impl.py
```diff
@@ -1003,32 +1013,90 @@ def _via_dtype(
...
+    cast_return=True,
 ) -> ndx.Array:
-    """Like ``_via_dtype`` but uses ``i64`` for integer or boolean types and ``float64``
-    for floating point types.
+    """Like ``_via_dtype`` but chooses a dtype from the available that doesn't result in
+    loss.
 
-    Raises TypeError if the provided arrays are neither.
+    Raises TypeError for non-numerical types and if the type can't be safely casted to
+    any available type.
```
cbourjau, line 1030:
```suggestion
    Raises
    ------
    TypeError
        For non-numerical types and if the type can't be safely casted to
        any available type.
```

ndonnx/_core/_impl.py
```diff
@@ -1003,32 +1013,90 @@ def _via_dtype(
...
+    elif isinstance(dtype, (dtypes.Integral, dtypes.NullableIntegral)) or dtype in (
         dtypes.nbool,
         dtypes.bool,
     ):
-        via_dtype = dtypes.int64
+        if int_dtype is not None and ndx.iinfo(dtype).bits <= ndx.iinfo(int_dtype).bits:
+            via_dtype = int_dtype
+        else:
+            raise TypeError(
+                f"Can't upcast signed type {dtype}. Available implementations are for {*available_types,}"
```
cbourjau, line 1064:
```suggestion
                f"Can't upcast signed type `{dtype}`. Available implementations are for `{*available_types,}`"
```

ndonnx/_core/_impl.py
```diff
@@ -1003,32 +1013,90 @@ def _via_dtype(
...
     elif isinstance(dtype, (dtypes.Floating, dtypes.NullableFloating)):
-        via_dtype = dtypes.float64
+        if (
+            float_dtype is not None
+            and ndx.finfo(dtype).bits <= ndx.finfo(float_dtype).bits
+        ):
+            via_dtype = float_dtype
+        else:
+            raise TypeError(
+                f"Can't upcast float type {dtype}. Available implementations are for {*available_types,}"
```
cbourjau, line 1074:
```suggestion
                f"Can't upcast float type `{dtype}`. Available implementations are for `{*available_types,}`"
```

ndonnx/_core/_impl.py
cbourjau: Using quotes indicates what may be grep'ed an what cannot.

cbourjau (commented):
Unfortunately, `float64` is usually the default in our use cases. I thought we had upstreamed these implementations in the past. Maybe @neNasko1 can find some time to add `float64` support for `ReduceProd` upstream? 

**Edit**: The related issue has a link to a discussion on the onnxruntime issue tracker. People seemed reluctant to implement ops. It should be an easy change though. I'd say we just make a PR upstream and see what happens.

ndonnx/_core/_impl.py
```diff
@@ -1003,32 +1013,93 @@ def _via_dtype(
...
+    for floating point types.
+
+    Raises TypeError if the provided arrays are neither.
+    """
+    return _via_upcast(
+        fn,
+        arrays,
+        int_dtype=dtypes.int64,
+        float_dtype=dtypes.float64,
+        use_unsafe_uint_cast=True,  # TODO this can cause overflow, we should set it to false and fix all uses
```
adityagoel4512, line 1098: Let's check this by triggering the Array API tests too.

adityagoel4512: We should automate the kernel support / standard diffing process - in the meantime, it [looks](https://github.com/Quantco/ndonnx/actions/runs/10076357986/job/27856564619#step:5:2484) like `Less` and `LessEqual` don't support unsigned types (but for some reason `Greater` and `GreaterEqual` do). There are only 13 uses so we can follow up on this.

neNasko1:
> Unfortunately, `float64` is usually the default in our use cases. I thought we had upstreamed these implementations in the past. Maybe @neNasko1 can find some time to add `float64` support for `ReduceProd` upstream?

I will submit a PR to upstream onnxruntime to see what their opinion on the matter is.

adityagoel4512 (approved):
Thanks, let's follow up on the unsigned cast safety in a separate PR.

neNasko1:
Should this be reverted in the next version of `onnxruntime` as per [PR](https://github.com/microsoft/onnxruntime/pull/21490#event-13730989070).

cbourjau:
Yes, let's revert them after the next release + a little bit of a grace period. I created an issue to keep track of it: https://github.com/Quantco/ndonnx/issues/42

PR #25, opened 2024-07-24
Allow `full` to work with scalar shapes.

adityagoel4512:
Currently, because the ONNX shape operator requires a 1D tensor as input, doing `ndx.full` with a scalar and lazy shape fails. This PR adds the extra dimension if the rank of the shape is 0. 

Note that we generally expect the rank to be known statically from ONNX shape inference and `ndx.Array.ndim` returns an int. You can construct specific cases where even the rank is data-dependent.

### DIFF
--- a/ndonnx/_funcs.py
+++ b/ndonnx/_funcs.py
@@ -144,10 +144,12 @@ def full(
         raise ValueError("fill_value must be a scalar")
 
     dtype = fill_value.dtype if dtype is None else dtype
-    shape = asarray(shape, dtype=dtypes.int64)._core()
-    return fill_value._transmute(lambda corearray: opx.expand(corearray, shape)).astype(
-        dtype
-    )
+    shape = asarray(shape, dtype=dtypes.int64)
+    if shape.ndim == 0:
+        shape = reshape(shape, [1])
+    return fill_value._transmute(
+        lambda corearray: opx.expand(corearray, shape._core())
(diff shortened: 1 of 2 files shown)


ndonnx/_core/_impl.py
```diff
@@ -890,6 +890,8 @@ def shape(self, x):
         current = x
         while isinstance(current, ndx.Array):
             current = next(iter(current._fields.values()))
+        if current.ndim == 0:
+            current = ndx.reshape(current, [1])
```
cbourjau, line 894: This looks fishy! A a scalar is not a rank-1 tensor with length 1. Furthermore, this fails under the following test:

```python
def test_shape():
    a = ndx.array(shape=(), dtype=ndx.int64)
    a.shape
```

Things pass if `a` is constant because the code paths are different.

cbourjau (changes requested):
See comment.

cbourjau:
As discussed, let's also reconsider if we have to go beyond the standard for `full`. The standard accepts `int | tuple[int, ...]` for the shape, but we also allow for arrays. Is this really necessary given the existence of `full_like`?

ndonnx/_core/_impl.py
adityagoel4512: Indeed, this slipped in. Thanks for pointing it out.

cbourjau (approved):

PR #27, opened 2024-07-26
Chores

cbourjau:
This PR contains various chores; none of them change the public API:

- We no longer raise DeprecaionWarnings that the user can do nothing about (follow up to #14). 
- While we still export the deprecated and public `promote_nullable` function I renamed its private conterpart to `into_nullable` which I find more appropriate.
- Tests are now found in `tests/` rather than in `tests/ndonnx/` This allows us to get rid of some schenanigans in `pyproject.toml`.
- Filtered out / addressed most warnings from the test suite. The last warning is fixed upstream in Spox https://github.com/Quantco/spox/pull/165


tests/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright (c) QuantCo 2023-2024
```
adityagoel4512, line 1: This seems dubious. This used to work without this before - can you explained what changed?

adityagoel4512, line 1: I'd prefer if you add back `pythonpath` in the pytest options and remove this `__init__.py`.

adityagoel4512 (changes requested):

adityagoel4512 (approved):

PR #28, opened 2024-07-26
Validate field dtypes explicitly

adityagoel4512:
Previously, we only validated that the field names are compatible with the dtype of the Array in the private `Array._from_fields`. We will now raise a more informative error if the dtypes are also incompatible - the user was expected to provide the correct thing previously in any case.

This isn't a public function and this error would only be reached in the case of custom dtype implementations deciding to access this private utility, so I've omitted it from the change log.


ndonnx/_array.py
```diff
@@ -88,13 +88,13 @@ def _from_fields(
...
-        ):
-            raise ValueError(
-                f"Expected fields {dtype._fields()} for StructType {dtype}, received {fields}"
+        field_dtypes = {name: field.dtype for name, field in fields.items()}
+        expected_dtypes = (
+            {"data": dtype} if isinstance(dtype, dtypes.CoreType) else dtype._fields()
+        )
+        if field_dtypes != expected_dtypes:
+            raise TypeError(
+                f"Expected fields {expected_dtypes} for {dtype}, received {fields}"
```
cbourjau, line 97:
```suggestion
                f"Expected fields `{expected_dtypes}` for `{dtype}`, received `{fields}`"
```

cbourjau (approved):

Issue #29, opened 2024-07-29
Track more recent and in-tree version of array-api-tests


cbourjau:
We are currently tracking an old out-of-tree [commit](https://github.com/data-apis/array-api-tests/tree/ea15d5792bbaeb57d2149e74870ad814730df704) of the array-api-tests. Let's move to a more recent commit that is part of upstream's `main`.

PR #32, opened 2024-07-29
Add complex data types

cbourjau:
Updating array-api-tests fails because the newer test suite crashes when looking for complex types. This PR adds complex types even though they appear to be totally broken with respect to support in the onnx package itself. I'm sure there is a better way to deal with this situation and I'm happy to hear your opinions! 

Related to #29


rgommers:
If complex dtypes are broken in ONNX, I think the test suite shouldn't compel you to add them. There is https://github.com/data-apis/array-api-tests?tab=readme-ov-file#skipping-dtypes, which I think should work for skipping complex dtypes. If not, maybe that requires an upstream fix.

cbourjau:
Thanks for dropping in @rgommers ! Unfortunately, the `array-api-tests` [call](https://github.com/data-apis/array-api-tests/blob/827edd804bcace9d64176b8115138d29ae3e8dec/array_api_tests/dtype_helpers.py#L326) `ndx.asarray(complex()).dtype` in the collection phase regardless whether complex types are meant to be skipped which causes a [crash](https://github.com/Quantco/ndonnx/actions/runs/10162286025/job/28102636353#step:5:18) (see this [branch](https://github.com/Quantco/ndonnx/tree/update-api-tests-without-complex)).

adityagoel4512:
Thanks for this. A few points:

- Your test adding two complex numbers together fails due to ONNX shape inference failing when trying to infer the types for the cast operator:

  ```
  E           RuntimeError: [ShapeInferenceError] (op_type:Cast, node name: _this_): input typestr: T1, has unsupported type: tensor(complex128)
  
  .pixi/envs/default/lib/python3.12/site-packages/onnx/shape_inference.py:46: RuntimeError
  ```
  
  As noted in the error, it is clear that the ONNX standard does not support casting of complex numbers - see https://github.com/onnx/onnx/blob/main/docs/Operators.md#type-constraints-21.
  
  The reason any casting is occurring in the first place is because the add implementation for core types ( `complex64/complex128` is one since you subclass `Floating`) first performs type promotion. This is something that will be resolved as part of https://github.com/Quantco/ndonnx/pull/17 systematically. You shouldn't need to support casting to implement addition.

- The `CoreOperationsImpl` that contains implementations of functionality for the core dtypes exposed by ndonnx may not be ready for complex numbers. If we want to add functionality bit by bit, it is worth adding a new implementation block for complex numbers as support in ONNX seems sparse. We will get better error messages. For instance not implementing `add` gives:

```
TypeError: Unsupported operand type(s) for add: 'Complex128' and 'Complex128'
```

This is more readable than a long ONNX stack trace.

cbourjau:
I'll close this PR given:
 1. Complex types are not particularly useful given the current state of the ONNX standard
 2. It seems like we are well able to run the array-api-tests without them (#37 )

PR #33, opened 2024-07-29
Update array-api-tests and run against newer version

cbourjau:
This PR updates the array-api-tests submodule and tests against a more recent version of the standard. It requires the complex data types to load (#32 ). The latest test suite gives a decent amount of failures to work through.


adityagoel4512:
Thanks! Could you add the failures to `skips.txt`. The diff here will show us what to work through clearly and allow this PR to be merged.

cbourjau:
Let's first figure out if we can get away with not adding complex types in #32 . It does seem easy enough to patch the `array-api-tests` on second sight. I'll mark this as "draft" in the meantime.

Issue #36, opened 2024-07-30
Split out string function implementations


adityagoel4512:
Strings are monolithically shoehorned into `CoreOperationsImpl` for historical reasons even though this mainly deals with numeric dtypes. We should split these out since some cases slipped through. Consider:

```
>>> x = ndx.array(dtype=ndx.utf8, shape=("M",))
>>> y = ndx.sin(x)
```

This is illogical and should error out as having been not implemented. Instead we can obtain the following ONNX graph with implicit casts inserted.

![modelsin](https://github.com/user-attachments/assets/f6ccb3cb-b113-4965-8da5-2097c5e74e29)

PR #40, opened 2024-08-01
Dispatch function implementation to dtype

adityagoel4512:
This PR splits off the implementation of top-level operations for the string and boolean dtypes from the numerical dtypes. Previously, this was all more or less monolithically together in `_impl.py` for legacy reasons. With this refactoring, we gain a few things:

- We get clean errors more consistently when operations don't make any sense. For instance ``ndx.sin(ndx.asarray(["a", "b"]))`` now gives the error message: ``UnsupportedOperationError: Unsupported operand type for sin: 'Utf8'`` which slipped through the cracks (closes https://github.com/Quantco/ndonnx/issues/36).
- It reduces the complexity involved with ensuring we dispatch to the right ONNX operator.

In making this more explicit, it also becomes apparent that especially the creation functions could not be implemented by custom dtypes. While this may seem unrelated, it follows pretty directly on the theme of dispatching via the dtype. Happy to split a new PR if it's needed for reviewing purposes although I would try and review the two together if it is reasonable to do so (I can help).

In a subsequent PR, we will reduce the complexities when interoperating user-defined dtypes with dtypes exported by ndonnx stemming from type-promotion etc by validating the input dtypes more systematically.


ndonnx/_core/_numericimpl.py
```diff
@@ -16,99 +18,88 @@
...
     from ndonnx._corearray import _CoreArray
 
 
-class CoreOperationsImpl(OperationsBlock):
+class NumericOperationsImpl(OperationsBlock):
     # elementwise.py
 
     def abs(self, x):
-        return _unary_op(x, opx.abs)
+        return unary_op(x, opx.abs)
```
cbourjau, line 40: I realize there is still the eager propagation behind this function, but is there anything conceptually wrong with doing:
```python
        import spox.opset.ai.onnx.v21 as op
        return ndx.from_spox_var(op.abs(x.spox_var()))
```

This would feel correct to me as a default. If we want to do any dtype specific logic, it should be implemented in dtype-specific subclass of `NumericOperationsImpl`, shouldn't it? Meanwhile, I wonder if we can't delegate the eager propagation up the stack (closer) to the public interface.

ndonnx/_core/_numericimpl.py
```diff
@@ -357,14 +291,14 @@ def argmax(self, x, axis=None, keepdims=False):
         if axis is None:
             reshaped_x = ndx.reshape(x, [-1])._core()
             if keepdims:
-                return _from_corearray(
+                return from_corearray(
```
cbourjau, line 294: Similar to my previous comment, I'm can't quite see why we are taking the detour via `_CoreArray` and `opx`. Why not directly call into the Spox operators by default?

ndonnx/_core/_numericimpl.py
```diff
@@ -899,186 +745,124 @@ def shape(self, x):
...
-    out_dtype = args[0].dtype
-    if not isinstance(out_dtype, (dtypes.CoreType, dtypes._NullableCore)):
-        raise TypeError(
-            f"Expected ndx.Array with CoreType or NullableCoreType, got {args[0].dtype}"
+        return from_corearray(opx.shape(current))
+
+    def can_cast(self, from_, to) -> bool:
+        if isinstance(from_, dtypes.CoreType) and isinstance(to, ndx.CoreType):
+            return np.can_cast(from_.to_numpy_dtype(), to.to_numpy_dtype())
+        return NotImplemented
```
cbourjau, line 753: Just double checking: This is indeed what we want rather than returning `False` by default?

ndonnx/_core/_stringimpl.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class StringOperationsImpl(OperationsBlock):
+    def add(self, x, y) -> Array:
+        return binary_op(x, y, opx.string_concat)
+
+    def equal(self, x, y) -> Array:
+        return binary_op(x, y, opx.equal)
+
+    def can_cast(self, from_, to) -> bool:
```
cbourjau, line 28: Isn't this a reasonable default implementation?

ndonnx/_core/_stringimpl.py
```diff
@@ -0,0 +1,50 @@
...
+            return np.can_cast(from_.to_numpy_dtype(), to.to_numpy_dtype())
+        return NotImplemented
+
+    def zeros(
+        self,
+        shape,
+        dtype: dtypes.CoreType | dtypes.StructType | None = None,
+        device=None,
+    ):
+        return ndx.full(shape, "", dtype=dtype)
```
cbourjau, line 39: Why `ndx.` rather than `self.`? Also below.

ndonnx/_funcs.py
```diff
@@ -23,6 +20,10 @@
     promote,
 )
 
+
+class UnsupportedOperationError(TypeError): ...
```
cbourjau, line 24: Just checking: `TypeError` is what we want as a parent? Not `ValueError` or something else? This ends up being in the public API.

ndonnx/_core/_default.py
```diff
@@ -0,0 +1,594 @@
...
+        return NotImplemented
+
+    def asinh(self, x) -> ndx.Array:
+        return NotImplemented
+
+    def atan(self, x) -> ndx.Array:
+        return NotImplemented
+
+    def atan2(self, y, x) -> ndx.Array:
+        return ndx.atan(ndx.divide(y, x))
```
cbourjau, line 47: Is this meant as a class with actual "defaults" or is the only real default that we want `NotImplemented`? I feel the second is more reasonable.

ndonnx/_funcs.py
```diff
@@ -383,31 +387,12 @@ def equal(x, y):
...
-    raise TypeError(f"Unsupported operand type for {func_name}: '{x.dtype}'")
-
-
 def exp(x):
     return _unary("exp", x)
 
 
 def expm1(x):
-    return subtract(exp(x), 1)
+    return _unary("expm1", x)
```
cbourjau, line 395: The `OperationsBlock` `expm1` and all(?) other operations here. If it is not implemented it will/should return `NotImplemented`. Why do we still have the need to deal with stringy function names here?

ndonnx/_funcs.py
```diff
@@ -596,128 +578,46 @@ def numeric_like(x):
...
-    x = asarray(x)
-    if (out := x.dtype._ops.broadcast_to(x, shape)) is not NotImplemented:
-        return out
-    shape = asarray(shape, dtype=dtypes.int64)._core()
-    return x._transmute(lambda corearray: opx.expand(corearray, shape))
+    return x.dtype._ops.broadcast_to(x, shape)
 
 
 # TODO: onnxruntime doesn't work for 2 empty arrays of integer type
+# TODO: what is the appropriate strategy to dispatch? (iterate over the inputs and keep trying is reasonable but it can
```
cbourjau, line 585: Duplicate?

ndonnx/additional/_additional.py
```diff
@@ -139,8 +139,9 @@ def make_nullable(x: Array, null: Array) -> Array:
     TypeError
         If the data type of ``x`` does not have a nullable counterpart.
     """
-    x = ndx.asarray(x)
     out = x.dtype._ops.make_nullable(x, null)
     if out is NotImplemented:
-        raise TypeError(f"'make_nullable' not implemented for `{x.dtype}`")
+        raise ndx.UnsupportedOperationError(
```
cbourjau, line 144: What is the rule here? When do we return `NotImplemented` and when do we raise `UnsupportedOperationError`? Could you summarize again the decision why we don't always rely on raising from further down the stack?

cbourjau (commented):
I am primarily a bit confused about how deep the dispatching still is and why that needs to be. We currently have public-api -> `Dtype._ops` -> free functions dealing with `_CoreArray`s -> `_opset_extensions` -> Spox opset

Intuitively, I'd say we should be able more or less directly call into the spox opsets from the `_ops` members. Am I missing something?

Could you also summarize again why we want to return `NotImplemented` in some places rather than relying on raising something?

ndonnx/_core/_default.py
```diff
@@ -0,0 +1,594 @@
...
+    def shape(self, x) -> ndx.Array:
+        return NotImplemented
+
+    def fill_null(self, x, value) -> ndx.Array:
+        return NotImplemented
+
+    def make_nullable(self, x, null) -> ndx.Array:
+        return NotImplemented
+
+    def can_cast(self, from_, to) -> bool:
```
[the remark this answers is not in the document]
adityagoel4512: It's an Array API function: https://data-apis.org/array-api/latest/API_specification/generated/array_api.can_cast.html#can-cast.

ndonnx/_core/_default.py
```diff
@@ -0,0 +1,594 @@
...
+import ndonnx as ndx
+import ndonnx._data_types as dtypes
+import ndonnx._opset_extensions as opx
+
+from ._utils import (
+    from_corearray,
+)
+
+
+class OperationsBlock:
```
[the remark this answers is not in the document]
adityagoel4512: Not really sure, it has always been called this.

ndonnx/_core/_numericimpl.py
```diff
@@ -16,99 +18,88 @@
 import ndonnx._opset_extensions as opx
```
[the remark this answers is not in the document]
adityagoel4512: It has always been called this since the distant past (@neNasko1). There are some utilities there like `getitem` that aren't a single operator.

ndonnx/_core/_stringimpl.py
adityagoel4512: I'm a bit confused as to what you're referring to with "this".

ndonnx/_core/_stringimpl.py
adityagoel4512: Either is definitely fine.

adityagoel4512:
> Intuitively, I'd say we should be able more or less directly call into the spox opsets from the `_ops` members. Am I missing something?

If you ignore constant propagation then mostly yes. You could do that inline but it's a lot of code duplication. There are also some functions like `static_map` or `getitem` that do more than a single operator but I think your point is about the single operator ones.
 
> Could you also summarize again why we want to return `NotImplemented` in some places rather than relying on raising something?

We always want to return `NotImplemented` when we don't know what to do with a data type. `NotImplemented` is a Python [special value](https://docs.python.org/3/library/constants.html#NotImplemented) that indicates the operation is not implemented with respect to the type provided, used by the interpreter to go from `__add__` to `__radd__` etc for binary dunder methods. We repurpose this to route our implementations in a similar way. An exception would indicate a user error or something where we immediately break out of the current frame.

ndonnx/additional/_additional.py
adityagoel4512: See https://github.com/Quantco/ndonnx/pull/40#issuecomment-2275572788

ndonnx/_funcs.py
adityagoel4512: I suppose `ValueError` would be more accurate as the dtype is a value? I can change it, thanks.

ndonnx/_funcs.py
adityagoel4512: Sorry I don't fully follow the question. `_unary` will dispatch the operation and if it returns `NotImplemented`, it will raise the `UnsupportedOperationError`.

ndonnx/_funcs.py
adityagoel4512: The TODO? I think it's different. I'm unsure how to make `concat` custom dtype friendly due to the unsafeness of naively dispatching in the order of the arguments. Happy to take any suggestions :).

ndonnx/_core/_stringimpl.py
adityagoel4512: For `ndx.full` if you provide the dtype parameter then it will be that dtype that implements the `full` operation.

ndonnx/_core/_default.py
adityagoel4512: I agree, I'm going to break out the default shape implementations.

ndonnx/_core/_numericimpl.py
adityagoel4512: I'd say so. It would be a little strange if you implement casting between a custom type and another type and then this returned `False`.

ndonnx/_core/_interface.py
```diff
@@ -1,355 +0,0 @@
-# Copyright (c) QuantCo 2023-2024
```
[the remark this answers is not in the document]
adityagoel4512: I decided to change this back. `_interface.py` is strictly an interface now. There is `_shapeimpl.py` with default implementations of shape relation functionality.

ndonnx/_core/_numericimpl.py
adityagoel4512: > Meanwhile, I wonder if we can't delegate the eager propagation up the stack (closer) to the public interface.

I think we most certainly can move eager propagation up the stack (to the exposed function level I imagine like `ndx.add`). You would sacrifice the granularity at which your debugger can help you since stepping into the function wouldn't be super helpful, but it might be a valid trade off of discuss.

> If we want to do any dtype specific logic, it should be implemented in dtype-specific subclass of NumericOperationsImpl, shouldn't it?

Yes but can we revisit this in a subsequent PR? I think this is a separate item of work and wouldn't affect the public API.

ndonnx/_core/_numericimpl.py
adityagoel4512: As you mentioned there, we're doing eager propagation at that level of the call stack. I'd prefer to move broader discussion about this to its own GitHub Issue/offline if that's okay.

ndonnx/_core/_stringimpl.py
cbourjau: Yes, I'm referring to this implementation of `can_cast` which appears to be copied three times.

ndonnx/_funcs.py
cbourjau: Sorry, I don't know what happened to that original comment of mine. Please see the edit. I am wondering why we have to resort to passing stringy function names.

ndonnx/_core/_stringimpl.py
adityagoel4512: It's there 3 times since we have 3 disjoint data type implementations. It's not worth adding a superclass where it is implemented once just for this.

adityagoel4512: It's there 3 times since we have 3 disjoint data type implementations. It's not worth adding a superclass where it is implemented once just for this. I don't understand what you mean by "default" implementation?

ndonnx/_funcs.py
cbourjau: Yes, the TODO. Isn't this TODO referring to the exact same thing as the one in line 555?

ndonnx/_core/_stringimpl.py
cbourjau: I mean that instead of copying it three times we may want to put it [here](https://github.com/Quantco/ndonnx/blob/118ecdc19ba9e20f003ed4ca93412be9da22b679/ndonnx/_core/_interface.py#L411). It is even a reasonable default for a custom type, isn't it?

ndonnx/_funcs.py
adityagoel4512: Okay I understand the question now. I think something like this existed a while ago and so it's stuck around - I can only imagine it's because the public API is untyped so it was a way to be defensive at some point.

CHANGELOG.rst
```diff
@@ -12,14 +12,30 @@ Changelog
 **New feature**
```
cbourjau, line 12:
```suggestion
**New features**
```

CHANGELOG.rst
```diff
@@ -12,14 +12,30 @@ Changelog
...
 
 - Expose the :func:`ndonnx.isdtype` function.
-
+- Custom dtypes can now override array creation functions:
```
cbourjau, line 15:
```suggestion
- Custom data types can now override array creation functions:
```

CHANGELOG.rst
```diff
@@ -12,14 +12,30 @@ Changelog
...
+   - :func:`ndonnx.zeros_like`
+   - :func:`ndonnx.ones`
+   - :func:`ndonnx.ones_like`
+   - :func:`ndonnx.full`
+   - :func:`ndonnx.full_like`
+   - :func:`ndonnx.arange`
+   - :func:`ndonnx.arange`
+   - :func:`ndonnx.eye`
+   - :func:`ndonnx.tril`
+   - :func:`ndonnx.triu`
```
cbourjau, line 26: Technically, these are not creation functions.

CHANGELOG.rst
```diff
@@ -12,14 +12,30 @@ Changelog
...
+   - :func:`ndonnx.linspace`
+- Custom dtypes can now provide implementations for :func:`ndonnx.where`.
```
cbourjau, line 28:
```suggestion
- Custom data types can now provide implementations for :func:`ndonnx.where`.
```

CHANGELOG.rst
```diff
@@ -12,14 +12,30 @@ Changelog
...
+- The :class:`ndonnx._experimental.UniformShapeOperations` now provides implementations of shape operations that are generic across all data types where each constituent field has the same shape (that of the overall array).
 
 **Other changes**
 
 - Fixed various deprecation warnings.
+- Invoking a function using arrays with dtypes lacking a corresponding implementation now raise a :class:`UnsupportedOperationError`.
```
cbourjau, line 34:
```suggestion
- Invoking a function using arrays of data types that lack a corresponding implementation now raises an :class:`UnsupportedOperationError`.
```

cbourjau (approved):
Let's ship this for now, but I'm afraid this is not the end of the data type saga just yet.

ndonnx/_core/_stringimpl.py
adityagoel4512: Potentially but let's keep `_interface` actually just an interface rather than blessing a few functions with default implementations (it used to be like this and it's unclear what you get for free).

CHANGELOG.rst
adityagoel4512: "arrays of data types" should be "arrays with data types" in my opinion but I took the second half of the suggestion.

MatejUrbanQC:
Are you aware that this changes behaviour of some operations? For example
```Python
ndx.asarray(1) + ndx.asarray('a')
```
now results in an error
```
TypeError: Cannot promote non string scalar to string type
```
Before, it would succeed with the result `Array(1a, dtype=Utf8)`

To prevent this, we need some validation in the numeric impls that returns `NotImplemented` when it encounters a non-numeric type.

adityagoel4512:
Yes it also meant that you could do things like:

```
>>> ndx.sin(ndx.asarray(["1"]))
Array(['0.84147098'], dtype=Utf8)
```

I don't think your addition example was intentional behaviour either and was an example of the (accidental) implicit casting mentioned [here](https://github.com/Quantco/ndonnx/pull/40/files#diff-2c623f3c6a917be56c59d43279244996836262cb1e12d9d0786c9c49eef6b43cR39).

MatejUrbanQC:
I missed this part in the changelog

> `ndx.utf8` and `ndx.nutf8` cannot be promoted with any other dtypes.

It makes sense now

(I deleted the comment right before you replied)

Issue #38, opened 2024-08-01
Promotion of Python scalars is losing precission


cbourjau:
Type promotion currently tries to be smart by seemingly finding the [smallest fitting data type](https://github.com/Quantco/ndonnx/blob/634442344b577c3028e2819fde66612add0af4ff/ndonnx/_utility.py#L49). Unfortunately, this does not appear to be working as intended. The consequence is that we lose precision on Python scalars during the type promotion. The following tests case demonstrates the issue:

```python
def test_promote(val, strict):
    promoted, = ndx._utility.promote(val)
    promoted_np = promoted.to_numpy()

    np.testing.assert_equal(np.asarray(val), promoted_np, strict=strict)

test_promote(0.123456789, strict=False)
```
which fails with:
```
AssertionError: 
Arrays are not equal

Mismatched elements: 1 [/](http://localhost:8888/) 1 (100%)
Max absolute difference among violations: 2.04328156e-09
Max relative difference among violations: 1.65505805e-08
 ACTUAL: array(0.123457)
 DESIRED: array(0.123457, dtype=float32)
```
I.e., `0.123456789` has been demoted to a float32 in ndonnx while it was kept as a float64 in NumPy. The demotion did cost significant loss in precision.

adityagoel4512:
More context: https://github.com/numpy/numpy/issues/8207

PR #41, opened 2024-08-01
Fix promotion precision loss

adityagoel4512:
Closes https://github.com/Quantco/ndonnx/issues/38

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,7 +20,7 @@ Changelog
 
 **Bug fixes**
 
-- Fixes scalar promotion logic to more accurately reflect the Array API standard. Promotion requires at least one array to be present and scalars adopt the dtype of the arrays being promoted with it.
+- Fixes scalar promotion logic to more accurately reflect the Array API standard. Promotion requires at least one array to be present and scalars adopt the dtype of the arrays being promoted with it. `ndx.utf8` and `ndx.nutf8` cannot be promoted with any other dtypes.
 
 
 0.6.1 (2024-07-12)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,7 +20,7 @@ Changelog
 
 **Bug fixes**
 
-- Fixes bug causing python float scalars to be lost precision when being promoted.
+- Fixes scalar promotion logic to more accurately reflect the Array API standard. Promotion requires at least one array to be present and scalars adopt the dtype of the arrays being promoted with it.
 
 
 0.6.1 (2024-07-12)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,10 @@ Changelog
 
 - Fixed various deprecation warnings.
 
+**Bug fixes**
+
+- Fixes bug causing python float scalars to be lost precision when being promoted.
+
 
 0.6.1 (2024-07-12)
 ------------------
--- a/ndonnx/_utility.py
+++ b/ndonnx/_utility.py
@@ -56,8 +56,6 @@ def _promote_with_none(*args: Array | npt.ArrayLike) -> list[Array | None]:
             scalars.append(arr)
         elif isinstance(arg, (int, str, np.generic)):
             np_dtype = np.min_scalar_type(arg)
-            if np_dtype == np.dtype("float16"):
-                np_dtype = np.dtype("float32")
             arr = asarray(arg, dtypes.from_numpy_dtype(np_dtype))
             arr_or_none.append(arr)
             scalars.append(arr)
--- a/ndonnx/_utility.py
+++ b/ndonnx/_utility.py
@@ -14,7 +14,7 @@ from spox import Var
 # FIXME: Remove private import from Spox! Better to just use reshape!
 from spox._internal_op import unsafe_reshape
 
-import ndonnx._data_types as dtypes
+import ndonnx as ndx
 
 if TYPE_CHECKING:
     from ndonnx import Array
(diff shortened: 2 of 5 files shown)


tests/test_core.py
```diff
@@ -634,3 +634,11 @@ def test_array_creation_with_invalid_fields():
 def test_promote_nullable():
     with pytest.warns(DeprecationWarning):
         assert ndx.promote_nullable(np.int64) == ndx.nint64
+
+
+@pytest.mark.parametrize("val", [1, 1.0, 1.5, 0.123456789, "a"])
```
cbourjau, line 639:
```suggestion
@pytest.mark.parametrize("val", [1, 1.0, 1.5, 0.123456789, "a", np.float64(0.123456789)])
```

cbourjau: I think this will still be broken. `min_scalar_type` seems to have odd semantics.

tests/test_core.py
```diff
@@ -634,3 +634,11 @@ def test_array_creation_with_invalid_fields():
 def test_promote_nullable():
     with pytest.warns(DeprecationWarning):
         assert ndx.promote_nullable(np.int64) == ndx.nint64
+
+
+@pytest.mark.parametrize("val", [1, 1.0, 1.5, 0.123456789, "a", np.float64(0.123456789)])
```
cbourjau, line 639:
```suggestion
@pytest.mark.parametrize(
    "val", [1, 1.0, 1.5, 0.123456789, "a", np.float64(0.123456789)]
)
```

tests/test_core.py
```diff
@@ -634,3 +634,13 @@ def test_array_creation_with_invalid_fields():
...
+
+
+@pytest.mark.parametrize(
+    "val", [1, 1.0, 1.5, 0.123456789, "a", np.float64(0.123456789)]
+)
+def test_scalar_promote(val):
+    x = ndx.asarray([val] * 4)
+    actual = (x + val).to_numpy()
+    expected = x.to_numpy() + val
+    np.testing.assert_equal(actual, expected, strict=True)
```
cbourjau, line 646: `actual` and `expected` go through ndonnx. I think it would be better to test against NumPy directly:

```suggestion
@pytest.mark.parametrize(
    "val", [1, 1.0, 1.5, 0.123456789, "a", np.float64(0.123456789)]
)
def test_scalar_promote(val):
    promoted, = ndx._utility.promote(val)
    promoted_np = promoted.to_numpy()

    np.testing.assert_equal(promoted, promoted_np, strict=True)
```

adityagoel4512:
Can you please create a separate branch.

tests/test_core.py
adityagoel4512: Can you please create a separate branch. I don't quite understand your comment. This PR explicitly doesn't use `np.min_scalar_type` for python float scalars.

tests/test_core.py
adityagoel4512: `actual`  goes through ndonnx, while in `expected` the addition (and hence type promotion) happens in NumPy.

ndonnx/_utility.py
```diff
@@ -45,7 +45,16 @@ def _promote_with_none(*args: Array | npt.ArrayLike) -> list[Array | None]:
                 signed_integer = True
         elif isinstance(arg, np.ndarray):
             arr_or_none.append(asarray(arg))
-        elif isinstance(arg, (int, float, str, np.generic)):
+        elif isinstance(arg, float):
```
cbourjau, line 48: This currently works because some `np.generic`s such as `np.float64` are also instances of `float`. Is this deliberate? If so, how would we end up with a `float16` further below? How about we give up on the idea of implicitly finding the smallest dtype in favor of some simpler logic here?

tests/test_core.py
cbourjau: Turns out that `np.float64` objects are instances of `np.generic` and `float`. Thus `np.min_scalar_type` is indeed not used, event though I'd say it is not overly explicit. If it had been used, we would have seen the following sematics that I do find very surprising:
```
np.min_scalar_type(np.float64(0.123456789))
>>> dtype('float16')
```

tests/test_core.py
cbourjau: Oh sorry, the suggestion is very nonsensical! I meant:

```python
@pytest.mark.parametrize(
    "val", [1, 1.0, 1.5, 0.123456789, "a", np.float64(0.123456789)]
)
def test_scalar_promote(val):
    promoted, = ndx._utility.promote(val).to_numpy()
    np.testing.assert_equal(promoted, np.asarray(va), strict=True)
```

I.e. wouldn't it be nicer, in the spirit of unit testing, to explicitly call `promote` in the test? This would also remove any implicit assumptions to what may be happening to a `val` on its way through `ndx.asarray(val).to_numpy()`.

tests/test_core.py
adityagoel4512: Yeah this definitely warrants an upstream issue in NumPy.

ndonnx/_utility.py
adityagoel4512: I meant to delete the line from before with the `float16` handling as it is now dead code, thanks for point it out.

I don't think you can give up on this, it is tested quite a lot in the array-api-standard tests and the type promotion DAG seems to suggest you go to `float32` if it is valid to do so.

tests/test_core.py
adityagoel4512: Is that what we want though? Consider:

```python
(Pdb) np.asarray(1.).dtype
dtype('float64')
```

```python
(Pdb) (np.asarray(1., dtype=np.float32) + 1.).dtype
dtype('float32')
```

i.e. creation seems to go to the default int/float dtype (not the smallest dtype). Promotion with scalars is independent to this as far as I understand it.

cbourjau: Ah, I see! Thanks! From a testing perspective I still think we should call `promote` directly, since that is ultimately what we want to unit test here:
```python
promoted, = ndx._utility.promote(val).to_numpy()
```
But then we should probably explicitly check that `promoted` is what we want it to be.

ndonnx/_utility.py
cbourjau: Thanks for the link! The section on dealing with Python scalars seems very relevant and rather easy to follow. Furthermore, there is this statement which seem to be a clear argument against using `numpy.min_scalar_type`: 
> Only the types of participating arrays matter, not their values

adityagoel4512: It says the types of the **arrays** matter. For mixing arrays and [scalars](https://data-apis.org/array-api/latest/API_specification/type_promotion.html#mixing-arrays-with-python-scalars) it says that the scalar adopts the dtype of the array. This ends up happening here by taking the minimum dtype and promoting the arguments (and also works for `result_type`).

tests/test_core.py
adityagoel4512: Sure I don't mind either way to be honest. The previous way one could compare with NumPy.

tests/test_core.py
adityagoel4512: Yup totally, it is not obvious what is going on for floats inside `min_scalar_type` - more context is in https://github.com/numpy/numpy/issues/8207.

ndonnx/_utility.py
cbourjau: > For mixing arrays and [scalars](https://data-apis.org/array-api/latest/API_specification/type_promotion.html#mixing-arrays-with-python-scalars) it says that the scalar adopts the dtype of the array. This ends up happening here by taking the minimum dtype the scalar can represent and promoting the arguments

What you describe is not "adopting the data type of the array", though. Instead, the algorithm you describe may result in a data type depending on the value of the Python scalar, doesn't it. Consider a value such as `1.1` which is more precisely represented in a `float64` than a `float32`. Further, consider an array of data type `float32`. As I read the standard, the promotion of the two should result in `float32` not `float64`. This is also what NumPy 2 is doing:

```python
import numpy as np
assert np.__version__ == "2.0.0"

val = 1.1
assert np.float32(val) != np.float64(val)

arr = np.array(1.0, np.float32) + val
assert arr.dtype == np.float32
```

Ultimately, I suppose our problem may boil down to the fact that type promotion is only defined by the standard if at least one argument is actually a standard `Array`.

adityagoel4512: Fair enough. It's interesting that the precision of the python scalar ends up having no relevance (and so you can lose precision like in the example), even though it is not an array yet.

CHANGELOG.rst
```diff
@@ -18,6 +18,10 @@ Changelog
 
 - Fixed various deprecation warnings.
 
+**Bug fixes**
+
+- Fixes bug causing python float scalars to be lost precision when being promoted.
```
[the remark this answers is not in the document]
adityagoel4512: This PR has slightly changed as you can still lose precision when promoting with scalars - if this happens then the remedy is that the array you are promoting with must have the correct dtype.

cbourjau (approved):
Thanks!

ndonnx/_utility.py
adityagoel4512: I should have looked into this a bit more. I'm not sure this is entirely consistent and it leads to some fairly odd situations now:

```
(Pdb) np.asarray([1, 2]) + 2.32
array([3.32, 4.32])
(Pdb) ndx.asarray([1, 2]) + 2.32
Array([3 4], dtype=Int64)
```

cbourjau: Oh yes, that aint good! I suppose we are still not following the letter of the standard, though:

> ...
>   - a Python int or float for real-valued floating-point array data types.
>
> ...
> Provided the above requirements are met, the expected behavior is equivalent to:
>  1. Convert the scalar to zero-dimensional array with the same data type as that of the array used in the expression.
>  2. Execute the operation for array <op> 0-D array (or 0-D array <op> array if scalar was the left-hand argument).

cbourjau: Oh no, that ain't good! Let's discuss in the follow up PR.

PR #43, opened 2024-08-02
Validate core dtypes for core operation implementations

adityagoel4512:
This supersedes https://github.com/Quantco/ndonnx/pull/17. It allows you to, for instance, implement binary operators like `add` where one argument is a user-defined dtype array while another is a type exported by `ndonnx` without falling into promotion errors.


cbourjau (changes requested):
It appears that something went wrong when merging the base branch into `main`. Regardless of that, I am skeptical of implementing this validation via a class decorator. Did you consider implementing it in `ndonnx._core._utils.variadic_op` instead?

adityagoel4512:
> Did you consider implementing it in ndonnx._core._utils.variadic_op instead?

The function implementations are much easier to reason about if you can assume that the data types of the array are known. e.g. you can type promote safely, etc. We can make the validation function decorators but I'd want to know more about the reasons for your skepticism first.

cbourjau:
The class decorator is very implicit and lives outside of the usual inheritance. What happens with subclasses? Should they be decorated again? What should be the behavior for private helper functions?  I realize that the data type abstraction is broken more generally and that the need for this work-around is a symptom rather than the place to properly fix it. If we can't find a nicer work-around right now so be it, but this is technical debt that really needs a proper fix soon.

adityagoel4512:
We can make them function decorators to address the subclass point but naturally any design work on the dtypes is outside the scope of this PR and requires some separate thinking.

cbourjau:
I reworded my previous comment. I think using individual function decorators is maybe a bit excessive boiler plate. If you are happy with this solution then I think we can merge this!

cbourjau (approved):

PR #44, opened 2024-08-02
Add tests for concat with empty array

MatejUrbanQC:
The shape inference fails when inputs are eager. Added a xfail test case for it.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -634,3 +634,39 @@ def test_array_creation_with_invalid_fields():
 def test_promote_nullable():
     with pytest.warns(DeprecationWarning):
         assert ndx.promote_nullable(np.int64) == ndx.nint64
+
+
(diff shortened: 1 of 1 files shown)


tests/test_core.py
```diff
@@ -634,3 +634,39 @@ def test_array_creation_with_invalid_fields():
 def test_promote_nullable():
     with pytest.warns(DeprecationWarning):
         assert ndx.promote_nullable(np.int64) == ndx.nint64
+
+
+@pytest.mark.xfail
```
adityagoel4512, line 639:
```suggestion
@pytest.mark.xfail(reason="https://github.com/onnx/onnx/issues/6276")
```

adityagoel4512 (changes requested):

cbourjau (approved):

PR #47, opened 2024-08-12
Fix error message if full_like is not implemented

adityagoel4512:
Small mistake in the error message is printed if a dtype has not implemented `full_like`.

### DIFF
--- a/ndonnx/_funcs.py
+++ b/ndonnx/_funcs.py
@@ -137,7 +137,7 @@ def full_like(
     ) is not NotImplemented:
         return out
     raise UnsupportedOperationError(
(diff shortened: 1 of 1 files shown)


cbourjau (approved):

ndonnx/_funcs.py
```diff
@@ -137,7 +137,7 @@ def full_like(
     ) is not NotImplemented:
         return out
     raise UnsupportedOperationError(
-        f"Unsupported operand type for full_like: '{dtype}'"
+        f"Unsupported operand type for full_like: '{x.dtype}'"
```
cbourjau, line 140:
```suggestion
        f"Unsupported operand type for full_like: `{x.dtype}`"
```

PR #52, opened 2024-08-13
Fix `where` changing dtype of nullable arrays

MatejUrbanQC:
Before, when `where` got two identical nullable arrays, the type got changed to non-nullable.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -775,3 +775,27 @@ def test_cross_kind_promotions(x, y):
         y = ndx.asarray(y)
     onnx_result = x + y
     np.testing.assert_equal(onnx_result.to_numpy(), np_result)
+
+
+# np.ma.allequal doesn't work correctly with identical objects, so we use a lambda to create distinct objects
+# https://github.com/numpy/numpy/issues/27201
+@pytest.mark.parametrize(
+    "x",
+    [
+        lambda: ndx.asarray([1, 2, 3], dtype=ndx.int64),
+        lambda: ndx.asarray([1.0, 2.0, 3.0], dtype=ndx.float64),
+        lambda: ndx.asarray([1, 2, 3]).astype(ndx.nint64),
+        lambda: ndx.asarray([1.0, 2.0, 3.0]).astype(ndx.nfloat64),
+        lambda: ndx.asarray([]).astype(ndx.nfloat64),
+    ],
+)
+def test_where_constant(x):
+    a = x()
+    b = x()
+    cond = ndx.array(shape=(), dtype=ndx.bool)
+
+    result = ndx.where(cond, a, b)
+
+    expected = x()
+    assert result.dtype == expected.dtype
+    np.testing.assert_equal(result.to_numpy(), expected.to_numpy())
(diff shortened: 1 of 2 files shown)


tests/test_core.py
```diff
@@ -775,3 +775,27 @@ def test_cross_kind_promotions(x, y):
...
     onnx_result = x + y
     np.testing.assert_equal(onnx_result.to_numpy(), np_result)
+
+
+# np.ma.allequal doesn't work correctly with identical objects, so we use a lambda to create distinct objects
+# https://github.com/numpy/numpy/issues/27201
+@pytest.mark.parametrize(
+    "x",
+    [
+        lambda: ndx.asarray([1, 2, 3], dtype=ndx.int64),
```
cbourjau, line 785: It is okay to create such small arrays directly in the file scope. After all, the `lambda` has to live somewhere, too.

tests/test_core.py
```diff
@@ -775,3 +775,27 @@ def test_cross_kind_promotions(x, y):
...
+        lambda: ndx.asarray([1.0, 2.0, 3.0], dtype=ndx.float64),
+        lambda: ndx.asarray([1, 2, 3]).astype(ndx.nint64),
+        lambda: ndx.asarray([1.0, 2.0, 3.0]).astype(ndx.nfloat64),
+        lambda: ndx.asarray([]).astype(ndx.nfloat64),
+    ],
+)
+def test_where_constant(x):
+    a = x()
+    b = x()
```
cbourjau, line 794: Could you make this test more general? I.e. do the mask-nonmaks and nonmask-mask input configuration do what we want? Or is that tested elsewhere already? If you really only want to test the case for identical inputs then a more explicit test name would be nice.

tests/test_core.py
```diff
@@ -775,3 +775,27 @@ def test_cross_kind_promotions(x, y):
...
+    cond = ndx.array(shape=(), dtype=ndx.bool)
+
+    result = ndx.where(cond, a, b)
+
+    expected = x()
```
cbourjau, line 799: I had similar discussions with @adityagoel4512 in the past, but I'd parameterized with NumPy arrays rather than ndonnx arrays so that the "expected" case can be expressed without any ndonnx dependency. I.e.:

```python
x:  np.ma.MaskedArray = ...
ndx_x = ndx.asarray(x)
result = ndx.where(cond=ndx.array(shape=(), dtype=ndx.bool), ndx_x. ndx_x)
assert result.dtype == ndx_x.dtype
np.testing.assert_equal(x, expected.to_numpy())
```

If we were to parametrize over different inputs as remarked above we could compute `result` directly via `numpy.where`, too. Am I missing something?

cbourjau (changes requested):
Thanks! You continue being the master tester!

tests/test_core.py
MatejUrbanQC: Actually I'm not even sure about parametrizing using numpy array. We need to check if the dtypes of the result is correct. It's much less explicit when we convert to numpy - we not only need to check the dtype, but also the class, either both have to be `masked_arrray` or both `array`.

cbourjau (approved):

tests/test_core.py
MatejUrbanQC: That doesn't work because `x.copy()` doesn't make a deep copy

PR #53, opened 2024-08-19
Fix parsing of scalar string arrays

MatejUrbanQC:


cbourjau (approved):

adityagoel4512 (changes requested):
Thanks for the fix. Please add a change log entry.

MatejUrbanQC:
> Thanks for the fix. This looks good but please add a change log entry.

Should I make a 0.7.1 entry or 0.8.0 or where should I put it?

adityagoel4512:
> > Thanks for the fix. This looks good but please add a change log entry.
> 
> Should I make a 0.7.1 entry or 0.8.0 or where should I put it?

0.7.1 as it is a bug fix.

adityagoel4512 (approved):

neNasko1 (approved):

PR #56, opened 2024-08-21
Assert masks more 

adityagoel4512:
`numpy.testing.assert_array_equal` always compares true where the mask is set (see https://github.com/numpy/numpy/issues/14624#issuecomment-563499537). We would like to be stricter than that in our testing.

I also set `strict=True` which exposed a few minor dtype discrepancies in `sum` and `prod`. We cannot quite trim `skips.txt` until kernel support improves but this can happen when that is available.


tests/test_core.py
```diff
@@ -602,19 +618,19 @@ def test_truediv():
         ndx.nint16,
         ndx.nint32,
         ndx.nint64,
-        ndx.uint8,
```
cbourjau, line 605: Does the test fail for these unsigned data types? If so, maybe we could keep the test but mark it as xfail?

tests/utils.py
```diff
@@ -70,3 +70,17 @@ def get_numpy_array_api_namespace():
...
+def assert_array_equal(
+    actual: np.ndarray,
+    expected: np.ndarray,
+):
+    np.testing.assert_array_equal(
+        actual, expected, strict=np.__version__.startswith("2")
+    )
+    assert isinstance(expected, np.ma.masked_array) ^ (
+        not isinstance(actual, np.ma.masked_array)
+    )
```
cbourjau, line 84: We want to assert that either both are masked or both are not masked, do we not? I'd find the following easier to read:

```suggestion
    # Assert that either both or neither is masked
    assert isinstance(expected, np.ma.masked_array) == isinstance(actual, np.ma.masked_array)
```

cbourjau (approved):

tests/test_core.py
adityagoel4512: These fail because of lack of kernel support and there is the test for this here: https://github.com/Quantco/ndonnx/pull/56/files#diff-938de46e643ab091cff6a7c23e4e752e8907e2266f47889d988923352f7a1058R636-R650. We previously did not follow the correct dtype determination algorithm but we could also just make the default unsigned int type uint32.

adityagoel4512: I changed my mind about creating a breaking change in https://github.com/Quantco/ndonnx/pull/58.

Issue #35, opened 2024-07-30
ndx.Array.shape should return a tuple


cbourjau:
While playing with ndonnx and the array-API compliant classes of scikit-learn I ran into some issue. One of them was that libraries may use code such as `len(array.shape)` to get the rank. While there is an explicit `Array.ndim` property, the former is a perfectly fine way to get the same information, too. In other cases, it may be necessary to get the number of features assuming a shape of `(M, n_features)` where `n_features` is statically known by calling `array.shape[1]`. Neither example currently works. 

On the other hand, I realize that there are also use cases where we need the shape as a lazy array which is currently served by the `Array.shape` implemenation. We should maintain that functionality but under a different name in order to serve both use cases and to maintain compliance with the standard.

adityagoel4512:
It sounds reasonable to use ONNX shape inference to give a real tuple. The only complication here of course is that shape inference may give you some symbolic dimensions:

```
>>> ndx.array(shape=("N", "M"), dtype=ndx.utf8).spox_var()
<Var from spox.internal@0::Argument->arg of str[N][M]
```

To comply with the [standard](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.shape.html#shape), we could set all symbolic dimensions to `None`. It might be worth checking if in practice people are checking for `None` dimensions but this is out of scope for ndonnx of course.

The other challenge is that there are some situations where even the rank is unknowable statically. Consider the following:

```
>>> x = ndx.array(shape=(2, 3), dtype=ndx.utf8)
>>> y = ndx.array(shape=("M",), dtype=ndx.int64)
>>> ndx.broadcast_to(x, y).spox_var()
<Var from ai.onnx@13::Expand->output of str[...]>
```

We would need to now raise in such situations (i.e. when shape inference gives us a `None`).

We currently have the `ndonnx.additional.shape` function that always provides a `ndx.Array` and is the better way to go about obtaining the dynamic shape anyway. It also should be overridable for custom dtypes. To that end, should we offload this calculation to the dtype itself so user-defined types may implement this however they see fit?

cbourjau:
> To comply with the [standard](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.shape.html#shape), we could set all symbolic dimensions to None

[...]

PR #57, opened 2024-08-22
Add `static_shape`

adityagoel4512:
Possible solution to https://github.com/Quantco/ndonnx/issues/35

### DIFF
--- a/ndonnx/_array.py
+++ b/ndonnx/_array.py
@@ -355,23 +355,31 @@ class Array:
 
     def __int__(self) -> int:
         eager_value = self.to_numpy()
-        if eager_value is not None and eager_value.size == 1:
-            return int(eager_value)
-        else:
-            raise ValueError(f"Cannot convert Array of shape {self.shape} to an int")
+        if eager_value is None:
+            raise ValueError(
+                f"The lazy Array {self} cannot be converted to a scalar int"
+            )
+        if eager_value.size != 1:
+            raise ValueError(
+                f"Array {self} with more than one element cannot be converted to a scalar int"
+            )
+        return int(eager_value)
 
     def __bool__(self) -> bool:
         eager_value = self.to_numpy()
-        if eager_value is not None and eager_value.size == 1:
-            return bool(self.to_numpy())
-        else:
-            raise ValueError(f"Cannot convert Array of shape {self.shape} to a bool")
+        if eager_value is None:
+            raise ValueError(f"The truth value of lazy Array {self} is unknown")
+        if eager_value.size != 1:
+            raise ValueError(
+                f"The truth value of Array {self} with more than one element is ambiguous"
+            )
+        return bool(self.to_numpy())
 
     def __len__(self) -> int:
         if isinstance(self.shape[0], int):
             return self.shape[0]
         else:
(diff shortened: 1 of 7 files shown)


ndonnx/_array.py
```diff
@@ -371,8 +368,8 @@ def __bool__(self) -> bool:
             raise ValueError(f"Cannot convert Array of shape {self.shape} to a bool")
 
     def __len__(self) -> int:
-        if isinstance(eager_value := self.shape, tuple):
-            return eager_value[0]
+        if isinstance(self.shape[0], int):
+            return self.shape[0]
         else:
             raise ValueError(f"Cannot convert Array of shape {self.shape} to a length")
```
neNasko1, line 374: This error is not very coherent, can we change it?

ndonnx/_core/_shapeimpl.py
```diff
@@ -24,6 +25,15 @@ def shape(self, x):
             current = next(iter(current._fields.values()))
         return from_corearray(opx.shape(current))
 
+    def static_shape(self, x) -> tuple[int | None, ...]:
+        current = x
+        while isinstance(current, ndx.Array):
```
cbourjau, line 30:
```suggestion
        # FIXME: Implicit assumption that all fields represent the shape of `x` correctly
        while isinstance(current, ndx.Array):
```

ndonnx/additional/__init__.py
```diff
@@ -1,12 +1,13 @@
 # Copyright (c) QuantCo 2023-2024
 # SPDX-License-Identifier: BSD-3-Clause
 
-from ._additional import fill_null, make_nullable, isin, shape, static_map
+from ._additional import fill_null, make_nullable, isin, shape, static_map, static_shape
```
cbourjau, line 4: Do we need to export `static_shape` publicly? When would one use it rather than `Array.shape`?

cbourjau (commented):
Thanks for taking this on! Is this causing significant churn in our dependent projects?

ndonnx/_array.py
adityagoel4512: Good point, and neither are `__bool__` and `__int__` error messages.

ndonnx/_core/_shapeimpl.py
adityagoel4512: We are in `_shapeimpl.py` which is used in dtypes that satisfy that condition - see https://github.com/Quantco/ndonnx/blob/main/ndonnx/_core/_shapeimpl.py#L17-L19.

ndonnx/additional/__init__.py
adityagoel4512: I considered leaving it private but I think the convention of having all overridable functionality being standalone public functions is valuable when contributing custom dtypes. On second thought we can always make it public later so I'll make it private for this release and then revisit this later.

adityagoel4512:
> Is this causing significant churn in our dependent projects?

Nothing very significant. The only thing that changes is that doing ``ndx.asarray(x.shape)`` becomes ``nda.shape(x)``. This is already the case usually and technically we always had the liberty to return `None` dimensions.

cbourjau (approved):

PR #59, opened 2024-08-23
Forbid iteration over dynamic dimensions

cbourjau:
Running the scikit-learn test suite on the latest 0.8 release revealed an infinite loop when using arrays in for-loops. Namely, it turns out that the following code currently loops indefinitely:

```python
a = ndx.array(shape=(2,), dtype=ndx.int64)

for item in a:
    ...
```

Looping like this over a dimension with static size is perfectly safe and reasonable to do. This PR ensures that the iterations over dimension with static length terminate correctly. 

The issue is more delicate for an array with a dynamic length such as in the following example. 

```python
a_dyn = ndx.array(shape=("N",), dtype=ndx.int64)

for item in a_dyn:
    ...
```
While infinite generators are not uncommon in Python, they only make sense if the user breaks out of the loop in some other way. It seems rather unlikely that this is what the user has in mind when running the second example, though. 

Rather than leaving this footgun in place, this PR explicitly prohibits `__iter__`ating over arrays where the first dimension is dynamic. It also prohibits iterating over 0-d arrays which is in line with NumPy's behavior

### DIFF
--- a/tests/test_iter.py
+++ b/tests/test_iter.py
@@ -7,8 +7,7 @@ import ndonnx as ndx
 
 
 def test_iter_for_loop():
-    n = 5
-    a = ndx.array(shape=(n,), dtype=ndx.int64)
+    a = ndx.array(shape=(5,), dtype=ndx.int64)
 
     for i, el in enumerate(a):  # type: ignore
         if i > n:
@@ -29,6 +28,7 @@ def test_create_iterators(arr):
     it = iter(arr)
     el = next(it)
     assert el.ndim == arr.ndim - 1
+    assert el.shape == arr.shape[1:]
 
 
 def test_0d_not_iterable():
--- a/tests/test_iter.py
+++ b/tests/test_iter.py
@@ -7,9 +7,11 @@ import ndonnx as ndx
 
 
 def test_iter_for_loop():
-    a = ndx.array(shape=(5,), dtype=ndx.int64)
+    n = 5
+    a = ndx.array(shape=(n,), dtype=ndx.int64)
 
     for i, el in enumerate(a):  # type: ignore
+        assert isinstance(el, ndx.Array)
         if i > n:
             assert False, "Iterated past the number of elements"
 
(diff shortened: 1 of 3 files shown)


tests/test_iter.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def test_iter_for_loop():
+    n = 5
+    a = ndx.array(shape=(n,), dtype=ndx.int64)
+
+    for i, el in enumerate(a):  # type: ignore
+        if i > n:
+            assert False, "Iterated past the number of elements"
+
```
adityagoel4512, line 16: In the body of the loop can you also assert `isinstance(el, ndx.Array)`

tests/test_iter.py
```diff
@@ -0,0 +1,42 @@
...
+        ndx.asarray([[1], [2]]),
+        ndx.array(shape=(2,), dtype=ndx.int64),
+        ndx.array(shape=(2, 3), dtype=ndx.int64),
+        ndx.array(shape=(2, "N"), dtype=ndx.int64),
+    ],
+)
+def test_create_iterators(arr):
+    it = iter(arr)
+    el = next(it)
+    assert el.ndim == arr.ndim - 1
```
adityagoel4512, line 31:
```suggestion
    assert el.ndim == arr.ndim - 1
    assert el.shape == arr.shape[1:]
```

tests/test_iter.py
```diff
@@ -0,0 +1,42 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+import pytest
+
+import ndonnx as ndx
+
+
+def test_iter_for_loop():
+    n = 5
+    a = ndx.array(shape=(n,), dtype=ndx.int64)
```
adityagoel4512, line 11:
```suggestion
    a = ndx.array(shape=(5,), dtype=ndx.int64)
```

adityagoel4512 (changes requested):

tests/test_iter.py
cbourjau: We use `n` in the assert in the loop below, too.

adityagoel4512 (approved):

PR #64, opened 2024-08-27
Fix make_nullable

MatejUrbanQC:
Use broadcast instead of reshape on the null array

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ Changelog
 **Bug fixes**
 
 - Various operations that depend on the array's shape have been updated to work correctly with lazy arrays.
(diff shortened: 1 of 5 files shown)


MatejUrbanQC:
I don't like that the implementation of `make_nullable` is repeated 3 times. I think we should move it to some "common implementations" block

adityagoel4512:
> I don't like that the implementation of `make_nullable` is repeated 3 times. I think we should move it to some "common implementations" block

I also noticed this - it ended up becoming a part of https://github.com/Quantco/ndonnx/pull/60/files#diff-1c3abe1838800ebe1bddf13cdf05b325b1884781c43f7e6f547950bf4e7f7dabR72.

adityagoel4512 (approved):

PR #70, opened 2024-08-29
Validate `dtype`-s in `where` to avoid exception

MatejUrbanQC:
This enables custom type overrides to operate with one argument being a core type.


adityagoel4512 (approved):
This is ready to go to me @MatejUrbanQC. Any final thoughts/concerns?

neNasko1:
LGTM as well.

neNasko1 (approved):

MatejUrbanQC:
I don't like that there is an implicit dependency of `CoreOperationsImpl` on `UniformShapeOperations` in the `where` operation. It would be explicit if `CoreOperationsImpl` would inherit from `UniformShapeOperations` (we would need to change the inheritance structure slightly)

adityagoel4512:
> We could make it explicit by CoreOperationsImpl inheriting from UniformShapeOperations (we would need to change the inheritance structure slightly)

https://github.com/Quantco/ndonnx/pull/70/commits/434d3903c9f59f08435017619c4c266654812bb8 does precisely this?

adityagoel4512:
> I don't like that there is an implicit dependency of `CoreOperationsImpl` on `UniformShapeOperations` in the `where` operation.
> 
> We could make it explicit by `CoreOperationsImpl` inheriting from `UniformShapeOperations` (we would need to change the inheritance structure slightly)

Totally, good catch.

adityagoel4512:
> I don't like that there is an implicit dependency of `CoreOperationsImpl` on `UniformShapeOperations` in the `where` operation.
> 
> We could make it explicit by `CoreOperationsImpl` inheriting from `UniformShapeOperations` (we would need to change the inheritance structure slightly)

Totally, I don't think https://github.com/Quantco/ndonnx/pull/70/commits/434d3903c9f59f08435017619c4c266654812bb8 did quite what it set out to 😀.

PR #72, opened 2024-08-29
Fix cumsum with dtype specified

adityagoel4512:


ndonnx/_core/_numericimpl.py
```diff
@@ -568,18 +568,17 @@ def cumulative_sum(
...
             else:
-                out = x.astype(_determine_reduce_op_dtype(x, dtype, dtypes.int64))
+                return NotImplemented
+        elif dtype == dtypes.uint64 or dtype == dtypes.nuint64:
+            raise ndx.UnsupportedOperationError(
+                f"Unsupported dtype parameter for cumulative_sum {dtype} due to missing kernel support"
+            )
         else:
-            out = out.astype(dtype)
+            out = x.astype(_determine_reduce_op_dtype(x, None, dtypes.int64))
```
MatejUrbanQC, line 581: `maximum_unsigned_dtype=int64` (with a signed integer) seems hacky. Could be just the naming of the parameter

ndonnx/_core/_numericimpl.py
```diff
@@ -568,18 +568,17 @@ def cumulative_sum(
...
+                out = x.astype(dtypes.int64)
             else:
-                out = x.astype(_determine_reduce_op_dtype(x, dtype, dtypes.int64))
+                return NotImplemented
+        elif dtype == dtypes.uint64 or dtype == dtypes.nuint64:
+            raise ndx.UnsupportedOperationError(
+                f"Unsupported dtype parameter for cumulative_sum {dtype} due to missing kernel support"
+            )
         else:
-            out = out.astype(dtype)
```
cbourjau, line 582: [According](https://data-apis.org/array-api/draft/API_specification/generated/array_api.cumulative_sum.html#cumulative-sum) to the standard, the idea of the `dtype` argument is to apply the casting before the cumsum operator so that it can be used to guard against overflows. It seems that now the `dtype` value is only used after the operation. Am I reading this wrong?

adityagoel4512: I originally implemented it strictly in this way but we spoke offline about this and came to the conclusion that this is fine. Firstly, we are dealing only the integral types here and we always upcast to `Int64` (note that `UInt64` is not supported yet since conda-forge is only up to ORT 1.18.1) for kernel support reasons. We always upcast for kernel support reasons _anyway_ and the intention of this parameter is to guard against overflows. Upcasting the argument to the maximum sized integer, doing the cumulative sum and then down casting to the provided `dtype` parameter will never cause any new overflows over and beyond casting to `dtype` then summing.

adityagoel4512: I think there is an interpretation where we must overflow in precisely the same way, but then the `dtype` argument becomes quite meaningless until we get some new ORT releases through with expanded support.

adityagoel4512: I think there is an interpretation where we must overflow in precisely the same way, but then the `dtype` argument becomes quite meaningless until we get some new ORT releases through with expanded support. Also I'm not sure there is any actual standardisation around overflows so I'd treat it as UB.

ndonnx/_core/_numericimpl.py
adityagoel4512: I think it's just poor naming. I'll change it to `uint64` since it's all the same functionally.

cbourjau (approved):

PR #77, opened 2024-09-26
Minor drive-by fixes

cbourjau:
These are some very minor drive-by changes. We were stuck on an older version of mypy due to an upper bound.

The `build` package was renamed as described [here](https://conda-forge.org/news/2024/08/21/remove-build/).

### DIFF
--- a/ndonnx/_corearray.py
+++ b/ndonnx/_corearray.py
@@ -88,7 +88,7 @@ class _CoreArray:
             indexing_expressions = len(tuple(idx for idx in index if idx is not None))
             if self.ndim != indexing_expressions:
                 raise IndexError(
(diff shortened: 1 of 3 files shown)


adityagoel4512 (approved):
Thanks, not sure why pixi adds an upper-bound by default

jtilly:
> Thanks, not sure why pixi adds an upper-bound by default

I call it the poetry disease. See https://github.com/prefix-dev/pixi/issues/639 for a discussion and https://github.com/prefix-dev/pixi/pull/1516 for the introduction of a fix.

```
pixi config set pinning-strategy no-pin
```
or 
```
pixi config set pinning-strategy major
```
strike me as better options for a library.

PR #84, opened 2024-10-24
Fix argmax and argmin

adityagoel4512:
Previously, `argmax` and `argmin` inserted unnecessary casts since the `cast_return` flag was True. This fixes this; an example can be seen in the documentation update.

It also makes argmin actually argmin.


cbourjau (approved):
Thanks! Feel free to make a release today.

CHANGELOG.rst
```diff
@@ -7,6 +7,12 @@
 Changelog
 =========
 
+0.9.3 (unreleased)
```
adityagoel4512, line 10:
```suggestion
0.9.3 (2024-10-25)
```

PR #88, opened 2024-12-31
Run Array API tests weekly

neNasko1:
This test enables nightly tests on the Array API, which tests using the "max-examples" set to the default and a non-constant seed. 

The time the action runs is taken from the [last scheduled job](https://github.com/Quantco/ndonnx/actions/runs/10072907802/workflow).

### DIFF
--- a/.github/workflows/array-api.yml
+++ b/.github/workflows/array-api.yml
@@ -3,7 +3,7 @@ name: Array API coverage tests
 on:
   push: # Run on all branches
   schedule:
-    - cron: "0 0 * * *" # Runs daily at 00:00 UTC
+    - cron: "* * * * *" # Runs daily at 00:00 UTC
 
 # Automatically stop old builds on the same branch/PR
 concurrency:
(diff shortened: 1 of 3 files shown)


.github/workflows/array-api.yml
```diff
@@ -1,5 +1,9 @@
 name: Array API coverage tests
-on: [push]
+
+on:
+  push: # Run on all branches
+  schedule:
+    - cron: "0 8 * * *" # Runs daily at 00:00 UTC
```
adityagoel4512, line 6:
```suggestion
    - cron: "0 8 * * 1" # Runs weekly at 08:00 UTC on Monday
```

adityagoel4512: I think a weekly run is sufficient. If a periodic test run like this fails, it would be great if a GH issue could be opened (see https://github.com/Quantco/glum/blob/main/.github/workflows/daily.yml#L48-L69 for an example).

README.md
```diff
@@ -106,3 +106,9 @@ Run the tests with:
 ```bash
 pixi run arrayapitests
 ```
+
+Or for a faster feedback cycle, with a fixed seed:
+
+```bash
+pixi run arrayapitests --max-examples 16 --hypothesis-seed=0
```
adityagoel4512, line 113:
```suggestion
```

README.md
```diff
@@ -106,3 +106,9 @@ Run the tests with:
 ```bash
 pixi run arrayapitests
 ```
```
adityagoel4512, line 108:
```suggestion
```bash
pixi run arrayapitests --max-examples 16 --hypothesis-seed=0
```
```

adityagoel4512 (changes requested):

adityagoel4512 (approved):

PR #90, opened 2025-01-03
Trigger CI on PRs from forks

adityagoel4512:


neNasko1:
I think we should also change the ["ref"](https://github.com/Quantco/ndonnx/pull/90/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR53) attribute to the default? 

The actions [README](https://github.com/actions/checkout) says:
```yaml
    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
    ref: ''
``` 

[Keeping it](https://github.com/Quantco/spox/pull/200/commits/dfc7b7fa95bfa54175b729e48aa58494235540e4) the same caused problems for me in `spox`.

PR #95, opened 2025-01-07
Implement `tensordot` and `repeat`

neNasko1:
This PR adds implementations to [tensordot](https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.tensordot.html) and [repeat](https://data-apis.org/array-api/2023.12/API_specification/generated/array_api.repeat.html). 

Note they currently segfault on the latest onnx release, so the arrayapi tests are disabled. Coverage is provided instead on our side with some tests.


ndonnx/_core/_shapeimpl.py
```diff
@@ -66,6 +66,25 @@ def expand_dims(self, x, axis):
             )
         )
 
+    def repeat(self, x, repeats, axis):
```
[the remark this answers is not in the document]
adityagoel4512: For example:

> If axis is None, let N = prod(x.shape) and if repeats is an array, repeats must be broadcast compatible with the shape (N,) (i.e., be a one-dimensional array having shape (1,) or (N,)).

The rank can always be recovered, so checking if `repeats` is 1D is easy to do. (the rank strictly speaking can be lost if it is data-dependent in contrived cases, but you can recover it with reshapes if you know what you're doing)

Also very easy to check:

> If repeats is an array, the array must have an integer data type.

Some of this may possibly fail down the stack, but we aim to localize the error handling where possible so the error messages are precise.

ndonnx/_core/_shapeimpl.py
```diff
@@ -66,6 +66,25 @@ def expand_dims(self, x, axis):
...
+            # TODO: this case can be optimized by broadcasting and reshaping
+            repeats = ndx.asarray(repeats)
+
+        repeats = ndx.broadcast_to(repeats, ndx.reshape(x_shape[axis], [1]))
+        indices = ndx.searchsorted(
+            ndx.cumulative_sum(repeats).astype(ndx.uint64),
+            ndx.arange(ndx.sum(repeats)),
+            side="right",
+        )
+        return ndx.take(x, indices, axis=axis)
```
adityagoel4512, line 86: Nice trick! Not sure if it's what you already allude to in your TODO but I believe the cumulative sum search is only required for the case where you have different repeat values. If `repeat` is a python scalar, we can skip the very expensive `searchsorted`. We can revisit this in a future PR but it'll be useful to know how much performance we're giving up.

xfails.txt
```diff
@@ -1,3 +1,7 @@
+# segmentation fault: https://github.com/onnx/onnx/pull/6570
+array_api_tests/test_manipulation_functions.py::test_repeat
+array_api_tests/test_linalg.py::test_tensordot
```
adityagoel4512, line 3: I don't see what https://github.com/onnx/onnx/pull/6570 has to do with `tensordot`?

adityagoel4512 (changes requested):
I'll need to play around with the einsum a little more but here's a first pass.

xfails.txt
neNasko1: After taking a good look, it seems I was mistaken - the actual segmentation fault came from the `onnxruntime` failing on `einsum` with zero dimensional input. 

Will submit a PR later, but in the meantime I improved the testing of `tensordot` to include some parametrised tests.

tests/test_core.py
```diff
@@ -9,6 +9,9 @@
 import numpy as np
 import pytest
 import spox.opset.ai.onnx.v19 as op
+from hypothesis import given
```
cbourjau, line 12: Let's keep hypothesis out of the ndonnx test suite for the time being, please. Would you mind summarizing what this test has that is lacking upstream? Does it make sense to upstream it?

xfails.txt
```diff
@@ -1,3 +1,8 @@
+# segmentation fault: https://github.com/onnx/onnx/pull/6570
+array_api_tests/test_manipulation_functions.py::test_repeat
+# segmentation fault: due to onnxruntime not handling einsum with dim 0
+array_api_tests/test_linalg.py::test_tensordot
```
cbourjau, line 4: If these tests segfault then we should put them into skips.txt rather than here. That the array-api-tests are green nonetheless is an indication that something else is fishy. That feeling is confounded by the fact that they finish within 20s. Could you take a look, please?

xfails.txt
neNasko1: > failing on einsum with zero dimensional input
The problem is in a shape, which has a zero dimension, not a shape that is with zero dimensions. E.g. `(0, 0, 1, 1)` is segfaulting, `()` is not.

Sorry for not being clear!

tests/test_core.py
neNasko1: The upstream is not missing anything, since we need to skip the test however @adityagoel4512 suggested we test ourselves, as to be sure that tensordot works correctly.

adityagoel4512: I was commenting on the fact that a few edge cases weren't included in our tests. I don't think we need to go to property-based tests for this.

neNasko1: Removed the parametrised testing and included some more diligent testing.

tests/test_core.py
```diff
@@ -1038,3 +1038,83 @@ def test_argmaxmin_unsupported_kernels(func, x):
...
+        (np.arange(60).reshape(3, 4, 5), 3, 1),
+        (np.arange(60).reshape(3, 4, 5), 3, 2),
+        (np.arange(60).reshape(3, 4, 5), 3, None),
+        (np.arange(60).reshape(3, 4, 5), np.arange(3), 0),
+        (np.arange(60).reshape(3, 4, 5), np.arange(4), 1),
+        (np.arange(60).reshape(3, 4, 5), np.arange(5), 2),
+        (np.arange(60).reshape(3, 4, 5), np.arange(60), None),
+    ],
+)
+def test_repeat(a, repeats, axis, lazy_repeats):
```
adityagoel4512, line 1102: What is "lazy" about the `lazy_repeats` case here? Even when `repeats` is `int`, it's immediately wrapped in `ndx.asarray(repeats)` in your implementation of `repeat`.

adityagoel4512: For a test that exercises shape inference for `repeat`, you could construct an ONNX model using `ndx.build` with a lazy `repeat` input.

neNasko1: The idea of the test is to test that if we actually supply the argument as a python integer, we wrap it correctly. I rewrote the test to be cleaner.

adityagoel4512: @neNasko1 please rebase

skips.txt
```diff
@@ -2,3 +2,9 @@
 
 array_api_tests/test_manipulation_functions.py::test_roll
 array_api_tests/test_data_type_functions.py::test_broadcast_arrays
+
+# segmentation fault: https://github.com/onnx/onnx/pull/6570
+array_api_tests/test_manipulation_functions.py::test_repeat
+
+# segmentation fault: due to onnxruntime not handling einsum with dim 0
```
adityagoel4512, line 9:
```suggestion
# segmentation fault: https://github.com/microsoft/onnxruntime/pull/23379
```

tests/test_core.py
```diff
@@ -1045,6 +1045,88 @@ def test_argmaxmin_unsupported_kernels(func, x):
...
+    [
+        (np.arange(60).reshape(3, 4, 5), np.arange(60).reshape([3, 4, 5]), 0),
+        (np.arange(60).reshape(3, 4, 5), np.arange(60).reshape([3, 20]), None),
+        (np.arange(60).reshape(3, 4, 5), np.arange(60).reshape([2, 3, 2, 5]), None),
+    ],
+)
+def test_repeat_raises(a, repeats, axis):
+    with pytest.raises(ValueError):
+        ndx.repeat(ndx.asarray(a), repeats, axis=axis).to_numpy()
+
```
adityagoel4512, line 1129:
```suggestion


```

neNasko1:
Thanks for merging into main!

adityagoel4512 (approved):

PR #97, opened 2025-01-16
Start using ORT 1.20 kernels

adityagoel4512:
This PR opens the door for using new kernels present in onnxruntime 1.20. We begin with `argmax` and `argmin`.


tests/test_core.py
```diff
@@ -1006,9 +1005,10 @@ def test_no_unsafe_cumulative_sum_cast():
         np.array([1, 2, 3, 4, 5], dtype=np.int8),
         np.array([1, 2, 3, 4, 5], dtype=np.int16),
         np.array([1, 2, 3, 4, 5], dtype=np.int32),
-        np.array([1, 2, 3, 4, 5], dtype=np.int32),
+        np.array([1, 2, 3, 4, 5], dtype=np.int64),
         np.array([1, 2, 3, 4, 5], dtype=np.uint8),
         np.array([1, 2, 3, 4, 5], dtype=np.uint16),
+        np.array([1, 2, 3, 4, 5], dtype=np.uint32),
         # (np.argmin, np.array([1, 2, 3, 4, 5], dtype=np.uint64)), -> onnxruntime
```
neNasko1, line 1012: Should this test be enabled as well?

neNasko1 (approved):
LGTM!

Just check whether the test I highlighted needs to be enabled.

tests/test_core.py
adityagoel4512: Don't think so as ArgMax and ArgMin don't have u64 kernels: https://github.com/microsoft/onnxruntime/blob/main/docs/OperatorKernels.md. We could do a lossy f64 cast but we should aim to upstream the kernel instead.

Issue #6, opened 2024-06-16
v2023.12 Array API compliance 


adityagoel4512:
(Ported this issue over since there are a few outstanding tasks).

The v2023.12 iteration of the Array API standard was recently released ([changelog](https://data-apis.org/array-api/latest/changelog.html)).

The following changes to ndonnx will be required:
- [ ] Add [__array_namespace_info__](https://data-apis.org/array-api/latest/API_specification/generated/array_api.info.__array_namespace_info__.html#array-namespace-info)
- [x] Move [clip](https://data-apis.org/array-api/latest/API_specification/generated/array_api.clip.html#clip) from additional namespace to main one.
- [x] Move cumsum from additional namespace to the main one and rename to [cumulative_sum](https://data-apis.org/array-api/latest/API_specification/generated/array_api.cumulative_sum.html#cumulative-sum).
- [x] Move [searchsorted](https://data-apis.org/array-api/latest/API_specification/generated/array_api.searchsorted.html#searchsorted) from additional namespace to main one.
- [x] Add [maximum](https://data-apis.org/array-api/latest/API_specification/generated/array_api.maximum.html).
- [x] Add [minimum](https://data-apis.org/array-api/latest/API_specification/generated/array_api.minimum.html).
- [ ] Add [signbit](https://data-apis.org/array-api/latest/API_specification/generated/array_api.signbit.html).
- [ ] Add [unstack](https://data-apis.org/array-api/latest/API_specification/generated/array_api.unstack.html).
- [ ] Add [repeat](https://data-apis.org/array-api/latest/API_specification/generated/array_api.repeat.html).
- [ ] Add [moveaxis](https://data-apis.org/array-api/latest/API_specification/generated/array_api.moveaxis.html).
- [ ] Add [tile](https://data-apis.org/array-api/latest/API_specification/generated/array_api.tile.html#tile).
- [ ] Add [hypot](https://data-apis.org/array-api/latest/API_specification/generated/array_api.hypot.html).

It looks like ndonnx already satisfies the clarifications to the specification, including the most urgent ones motivated by the ONNX use case in the first place like raising in intrinsically eager functions like __bool__.

The "additional" functions can be deprecated. Either functions are now part of the Array API or functions like ndindex, fill_null, etc that were highly motivated by dataframe applications can be defined by downstream projects ([onnx_pandas_compat](https://github.com/Quantco/onnx_pandas_compat) in practice).

cbourjau:
[...]

PR #101, opened 2025-01-24
(Array API support): Add `__array_namespace_info__` and `device`

adityagoel4512:
Continues with https://github.com/Quantco/ndonnx/issues/6.


ndonnx/_info.py
```diff
@@ -0,0 +1,70 @@
...
+    def devices(self) -> list:
+        return [ndonnx_device]
+
+    def dtypes(
+        self, *, device=None, kind: str | tuple[str, ...] | None = None
+    ) -> dict[str, ndx.CoreType]:
+        # We don't care for device since we are writing ONNX graphs.
+        # We would rather not give users the impression that their arrays
+        # are tied to a specific device when serializing an ONNX graph as
+        # such a concept does not exist in the ONNX standard.
```
cbourjau, line 46:
```suggestion
```
This comment would be better placed in the `NdonnxDevice` class, but we can forego it completely. It is somewhat self-explaining, isn't it?

ndonnx/_info.py
```diff
@@ -0,0 +1,70 @@
...
+        device=None,
+    ) -> dict[str, ndx.CoreType | None]:
+        # See comment in `dtypes` method regarding device.
+        return {
+            "real floating": ndx.float64,
+            "integral": ndx.int64,
+            "indexing": ndx.int64,
+            # We don't support complex numbers yet due to immaturity in the ONNX ecoystem, so "complex floating" is meaningless.
+            # The Array API standard requires this key to be present so we set it to None.
+            "complex floating": None,
```
cbourjau, line 65: Returning something wrong is still not compliant with the standard, but it might lead to issues that are harder to debug than what one would get if the key were just missing. I think we should rather not return `"complex floating"` as a key.

ndonnx/_data_types/aliases.py
```diff
@@ -55,3 +61,52 @@
 nuint32: NUInt32 = NUInt32()
 nuint64: NUInt64 = NUInt64()
 nutf8: NUtf8 = NUtf8()
+
+
+_canonical_names = {
```
cbourjau, line 66: I think `_canonical_names` and _kinds` should rather life with the `DType` objects, but that can be something for the future.

cbourjau (changes requested):

ndonnx/_info.py
adityagoel4512: > It is somewhat self-explaining, isn't it?

I think it's self-explanatory for us but I meant it as a note for other contributors since it's an unconventional implementation choice.

ndonnx/_info.py
adityagoel4512: I do agree however the tests as they stand explicitly assert all the keys are present. See https://github.com/data-apis/array-api-tests/pull/337.

ndonnx/_array.py
```diff
@@ -254,6 +254,19 @@ def shape(self) -> tuple[int | None, ...]:
...
             return static_shape(self)
 
+    @property
+    def device(self):
+        return device
+
+    def to_device(
+        self, device: _Device, /, *, stream: int | Any | None = None
+    ) -> Array:
+        if device is not device:
```
cbourjau, line 264:
```suggestion
        if device is not _Device():
```

ndonnx/_info.py
```diff
@@ -0,0 +1,62 @@
...
+        ndx.int16,
+        ndx.int32,
+        ndx.int64,
+        ndx.uint8,
+        ndx.uint16,
+        ndx.uint32,
+        ndx.uint64,
+    ]
+
+    def capabilities(self) -> dict:
```
cbourjau, line 28:
```suggestion
    def capabilities(self) -> dict[str, bool]:
```

ndonnx/_info.py
```diff
@@ -0,0 +1,62 @@
...
+        return {
+            "boolean indexing": True,
+            "data-dependent shapes": True,
+        }
+
+    def default_device(self):
```
cbourjau, line 34:
```suggestion
    def default_device(self) -> _Device:
```

ndonnx/_info.py
```diff
@@ -0,0 +1,62 @@
...
+        return device
+
+    def devices(self) -> list:
```
cbourjau, line 37:
```suggestion
    def devices(self) -> list[_Device]:
```

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "array-api-tests"]
 	path = api-coverage-tests
-	url = git@github.com:data-apis/array-api-tests.git
+	url = git@github.com:adityagoel4512/array-api-tests.git
```
cbourjau, line 3: Upstream got merged, so we can get back onto `main`.

ndonnx/_array.py
```diff
@@ -254,6 +254,19 @@ def shape(self) -> tuple[int | None, ...]:
         else:
             return static_shape(self)
 
+    @property
+    def device(self):
```
cbourjau, line 258:
```suggestion
    def device(self) -> _Device:
```

cbourjau (approved):
Thanks, Looks good! I just have a few more minor suggestions.

PR #104, opened 2025-01-31
Address xfails and mislabelled kind

adityagoel4512:
I mislabelled the integral kind in https://github.com/Quantco/ndonnx/commit/039264da5587edc5a65ebc6ac2f7f0d0f19176de. It wasn't caught since we skipped a relevant test from upstream.


cbourjau (approved):
Thanks!

Issue #75, opened 2024-09-23
Naming of `Array.spox_var()` and `ndonnx.from_spox_var`


cbourjau:
We have had discussions recently about the naming of these functions. I believe there are two parts to consider here: the prefix and the suffix part.

Concerning the prefix, it might be valuable to harmonize the naming with the common `unwrap` prefix used in spox (e.g., `Var.unwrap_tensor`), which signals the fallibility nicely, IMHO. Concerning the suffix we may consider using only `spox` or `var` in the naming. Thus, I see the following options for the unwrap part:
- `unwrap_var`
- `unwrap_spox`
- `unwrap_spox_var`
and one of the following constructor functions:
- `from_var`
- `from_spox`
- `from_spox_var`

I'd be in favor of `unwrap_spox_var` and `from_spox_var`. Any thoughts, @adityagoel4512 @neNasko1

adityagoel4512:
The falliability of `to_spox_var` is documented in the docstring although I can see why you say it might be clearer in `unwrap_spox_var`. If the main benefit is to make the name line up with Spox then I don't think that is a good enough reason for a breaking change. The ndonnx-Spox interoperability is used in a lot of different places.

adityagoel4512:
We should take advantage of not being 1.x and do this. Let's not give ourselves a strong deadline to remove `spox_var` but `unwrap_spox_var` is good to me 👍.

Issue #74, opened 2024-09-12
Handling compatibility with ONNX Runtime


MatejUrbanQC:
The ONNX Runtime is missing support for certain data types in some operations. This is currently handled in NDOnnx by casting the operand to a supported type before and casting back after the operation.

The problem is that this generates an inefficient ONNX graph without the user being aware.

## Example

Here's an example code with `uint64`, for which ONNX Runtime doesn't implement `Add`.
In the "Implicit cast" case, I let NDOnnx do the conversion for each operation.
In the "Explicit cast" case, I just cast to `int64` at the beginning of the whole computation and cast back at the end.

```Python
import time
import ndonnx as ndx
import onnx
import onnxruntime as ort
import numpy as np

# BUILD

def compute(x):
    return x + x + x + x + x + x

x = ndx.array(shape=('N',), dtype=ndx.uint64)

# Implicit cast:
y1 = compute(x)

model1 = ndx.build({'x': x}, {'y': y1})
onnx.save(model1, "model1.onnx")

# Explicit cast:
y2 = compute(x.astype(ndx.int64)).astype(ndx.uint64)

model2 = ndx.build({'x': x}, {'y': y2})
onnx.save(model2, "model2.onnx")

# RUN

input1 = np.arange(100000000, dtype=np.uint64)

options1 = ort.SessionOptions()
options1.optimized_model_filepath = "optimized1.onnx"
session1 = ort.InferenceSession(model1.SerializeToString(), options1)
start = time.time()
session1.run(None, {'x': input1})
end = time.time()
print(f"Implicit cast: {end-start:.3f}s")

options2 = ort.SessionOptions()
options2.optimized_model_filepath = "optimized2.onnx"
session2 = ort.InferenceSession(model2.SerializeToString(), options2)
start = time.time()
session2.run(None, {'x': input1})
end = time.time()
print(f"Explicit cast: {end-start:.3f}s")
```
### Result:
```
Implicit cast: 0.634s
Explicit cast: 0.305s
```
### Models:
Implicit:
![optimized1 onnx](https://github.com/user-attachments/assets/ff14df55-bdb9-4884-81af-d320f3586c62)
Explicit:
![optimized2 onnx](https://github.com/user-attachments/assets/dbdf8908-71f7-4e37-a197-e20fb49fa078)

## Suggested changes

There are at least 4 ways to handle these casts:
- Cast Silent - the current behaviour
- Cast with Warn - raise a Warning each time we do this casting
[...]

PR #110, opened 2025-03-10
Refactor internals

cbourjau:
This PR is a large refactor of the internal workings of ndonnx. The newly introduced concepts are described in an explicit [file](https://github.com/Quantco/ndonnx/blob/typed-array/docs/refactoring.md).

Breaking changes are limited to functions not found in the array-api standard and can likely be reduced further.

Closes https://github.com/Quantco/ndonnx/issues/75.
Closes https://github.com/Quantco/ndonnx/issues/74.


ndonnx/_typed_array/masked_onnx.py
```diff
@@ -0,0 +1,881 @@
...
+            return asncoredata(data, mask).astype(self)
+        else:
+            return asncoredata(self._unmasked_dtype.__ndx_create__(val), None).astype(
+                self
+            )
+
+    def __str__(self) -> str:
+        return type(self).__name__.lower()
+
+    @abstractmethod
```
adityagoel4512, line 68: Why is this abstract? Is it not just `self._unmasked_dtype.__ndx_create__(...)` and the same mask logic for all the masked implementations here?

ndonnx/_typed_array/masked_onnx.py
```diff
@@ -0,0 +1,881 @@
...
+nuint32 = NUInt32()
+nuint64 = NUInt64()
+
+nutf8 = NUtf8()
+
+# Union types
+#
+# Union types are exhaustive and don't create ambiguities with respect to user-defined subtypes.
+# TODO: Rename
+NCoreIntegerDTypes = (
```
adityagoel4512, line 295: NPrimitiveIntegerDTypes?

ndonnx/_typed_array/masked_onnx.py
```diff
@@ -0,0 +1,881 @@
...
+
+    @data.setter
+    def data(self, data: onnx.TyArray):
+        if data.dtype != self.dtype._unmasked_dtype:
+            raise ValueError(
+                f"expected data of type `{self.dtype._unmasked_dtype}` found `{data.dtype}`"
+            )
+        self._data = data
+
+    def disassemble(self) -> dict[str, Var]:
```
adityagoel4512, line 389: The optional mask creates a subtle but important breaking change: inference handling code that previously assumed that there is a one-two-one, static relationship between data type and "fields" see this invariant violated. This means you can read the metadata from an ONNX model but that does NOT tell you everything about the way to split up your inputs for the ONNX backend being used. i.e. inspecting the metadata, I have no idea if the "null" field is present or not.

Not every user may be in the position to *also* inspect the input fields in the generated protobuf (now required). In any case, I think there are two aspects to address before merging this:

- This is a very subtle but undocumented breaking change. Is it truly so necessary that it gets incorporated along with a substantial refactoring of internal behaviour? I understand that there will be perf wins.
- If it is so urgent, can we continue to have this very nice property that the dtype schema/metadata is sufficient to know exactly how to pass things to your ONNX model?

ndonnx/_typed_array/object_dtype.py
```diff
@@ -0,0 +1,347 @@
...
+
+if TYPE_CHECKING:
+    from spox import Var
+    from typing_extensions import Self
+
+    from .._types import NestedSequence, OnnxShape, PyScalar
+    from .indexing import GetitemIndex, SetitemIndex
+
+
+# TODO: The name is unfortunate. We cover far less than the "object"
```
adityagoel4512, line 27: rm or move into test suite? Better in downstream library in my opinion, as it is too specific to pandas.

ndonnx/_array.py
```diff
@@ -1,614 +1,443 @@
...
+        return Array._from_tyarray(+self._tyarray)
 
-    def __eq__(self, other):
-        return type(other) is _Device
+    def __repr__(self) -> str:
+        value_repr = ", ".join(
+            [f"{k}: {v}" for k, v in self._tyarray.__ndx_value_repr__().items()]
+        )
+        shape = self._tyarray.shape
+        return f"array({value_repr}, shape={shape}, dtype={self.dtype})"
```
adityagoel4512, line 342: The new print formatting gets a little long for constant arrays. Can we remove the shape for these?

tests/test_dtypes.py
```diff
@@ -1,436 +0,0 @@
...
-        )
-
-    def shape(self, x) -> Array:
-        return nda.shape(x.endpoints)[:-1]
-
-    def static_shape(self, x) -> tuple[int | None, ...]:
-        return x.endpoints.shape[:-1]
-
-
-class List(StructType):
```
adityagoel4512, line 188: This test was very valuable for testing the limits of the dtype system. I'm sure it's implementable again - can you please bring back the `List` test and/or have some custom dtype tests that illustrate how dtypes get defined from outside of ndonnx. Right now, everything (categoricals, object dtype) has been inlined into the library which won't be the long term way to support niche dtypes without a clear numpy counterpart.

adityagoel4512 (changes requested):
Thanks. We've agreed the general design already a few times so I go into various other directions here. I'm mainly concerned about the conflation of some subtle breaking changes (and I'm sure I cannot catch every single one given the size of this PR), obvious changes, and internal refactoring. Many (not all) of the breaking changes feel uncorrelated with "Refactor internals" being the primary motivation.

ndonnx/__init__.py
```diff
@@ -124,218 +176,197 @@
...
-    "nint64",
-    "nuint8",
-    "nuint16",
-    "nuint32",
-    "nuint64",
-    "NullableNumerical",
-    "Numerical",
+    "DateTime64DType",
+    "TimeDelta64DType",
+    "object_dtype",
```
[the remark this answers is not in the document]
adityagoel4512: "This" being 'object_dtype'.

adityagoel4512:
> Breaking changes are limited to functions not found in the array-api standard and can likely be reduced further.

I also think enumerating the breaking changes (some of which I've commented on already) will help with this. Breaking the metadata, for instance, would be something that is just as unacceptable as not implementing Array API code as users' production inference handling code will depend on this. We can mint a `v2` metadata if anything necessary comes up!

docs/experimental/experimental.rst
```diff
@@ -1,52 +1,27 @@
 Experimental
 ============
```
adityagoel4512, line 2: Thanks for already taking care of parts of the docs :). The Spox Integration page looks unedited.

.github/workflows/array-api.yml
```diff
@@ -9,6 +9,7 @@ on:
   push:
     branches:
       - "main"
+      - "typed-array"
```
adityagoel4512, line 12:
```suggestion
    branches:
      - "*"
```
?

adityagoel4512:
I also think we should indeed ship datetime64, timedelta64 but not the categorical or "object" dtype. It would be good to remove the pandas dependency added to the test environment.

tests/schemas/Bool.json
```diff
@@ -0,0 +1,17 @@
+{
```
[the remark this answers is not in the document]
cbourjau: fixed

tests/test_build_utils.py
```diff
@@ -23,18 +24,88 @@
...
+    # We must not break backwards compatibility. We test every type we
+    # support that it keeps producing the same schema.
+    import json
+    from pathlib import Path
+
+    # These files should not be update automatically
+    fname = Path(__file__).parent / f"schemas/{dtype}.json"
+
+    # Only set to `True` temporarily and only if there was a
+    # deliberate update to the schema.
```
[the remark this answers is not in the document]
cbourjau: I backported this code to #111 . You are right that the schemas should not be updated. It might be that we add more schemas, though, or that one may want to regenerate the files for some other reasons. If nothing else it is good to see how the files were generated.

ndonnx/_typed_array/object_dtype.py
cbourjau: Move to the [`object-array`](https://github.com/Quantco/ndonnx/tree/object-array) branch for now.

docs/experimental/experimental.rst
```diff
@@ -55,106 +30,4 @@ If you are using your own custom operator with a onnxruntime-compatible implemen
 User-defined data types
```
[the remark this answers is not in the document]
cbourjau: I added a bare-bones description that should get people started.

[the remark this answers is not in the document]
adityagoel4512: Can you please flesh this out with a real example (can be easily shared with the test suite). It's not worth putting in the documentation at all if it's just a link to an inbuilt data type as that doesn't show how to extend the library.

ndonnx/_propagation.py
```diff
@@ -1,211 +1,22 @@
...
+from collections.abc import Callable
+from functools import wraps
+from typing import Any, TypeVar, cast
+from warnings import warn
 
-import warnings
-from collections import namedtuple
-from pathlib import Path
+warn(
+    "the 'ndonnx._propagation' module is deprecated and can be simply removed.",
```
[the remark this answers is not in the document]
adityagoel4512: Okay, can you fix the capitalization though?

ndonnx/_typed_array/funcs.py
```diff
@@ -0,0 +1,313 @@
...
+def astyarray(
+    val: PyScalar | np.ndarray | TyArrayBase | Var | Array | NestedSequence,
+    dtype: None | DType = None,
+) -> TyArrayBase: ...
+
+
+def astyarray(
+    val: PyScalar | np.ndarray | TyArrayBase | Var | Array | NestedSequence,
+    dtype: None | DType[TY_ARRAY_BASE] = None,
+) -> TyArrayBase:
```
adityagoel4512, line 94: We should aim to have acyclic dependencies. This can be achieved by having each "abstraction" level only dependent on items at the same level or lower-level. Here, this function shouldn't accept an `Array` but only a `TyArrayBase` and the lazy import goes away. The caller can handle unwrapping the `Array`.

ndonnx/_typed_array/funcs.py
```diff
@@ -0,0 +1,313 @@
...
+if TYPE_CHECKING:
+    from .. import Array
+    from .._dtypes import TY_ARRAY_BASE
+    from . import onnx
+
+
+def _infer_sequence(
+    val: NestedSequence,
+) -> DType:
+    from . import onnx
```
adityagoel4512, line 27: We should factor out whatever is causing this cyclic dependency and depend on it from`onnx` submodule (still think that's a confusing name) and from this one.

ndonnx/_typed_array/onnx.py
```diff
@@ -0,0 +1,2125 @@
...
+        # It seems that there is currently a bug(?) in the type/shape
+        # inference in ONNX which prohibits us from concatenating
+        # empty 1D int32 and int64 arrays (see test case). We therefor
+        # do some hacky special-casing here for those types and those
+        # types only. Other data types are fine.
+        #
+        # TODO: File upstream bug; this may also be what caused the
+        # segfaults in onnxruntime in the past!
+        if self.dtype in (int32, int64) and self.ndim == 1:
+            axis_ = axis if axis is None else axis
```
adityagoel4512, line 681: Why is this line producing `axis` regardless of branch.

adityagoel4512, line 681: Why is this line producing `axis` regardless of branch.

ndonnx/_typed_array/onnx.py
```diff
@@ -0,0 +1,2125 @@
...
+            )
+        if var.unwrap_tensor().shape is None:
+            raise ValueError("'var' has no shape information")
+        self._var = var
+
+    def __ndx_value_repr__(self) -> dict[str, str]:
+        try:
+            # TODO: should this really be "data"?
+            return {"data": str(self.unwrap_numpy().tolist())}
+        except ValueError:
```
adityagoel4512, line 304: This try-catch looks like it'll be buggy. Specifically, a `ValueError` inside `unwrap_numpy()` can also be from invalid data, _not just because there's no data_. In this file you can see this as an example: https://github.com/Quantco/ndonnx/pull/110/files#diff-7e2260bdac28e59ef26d7f8e6f2f350474002467a5107dbcaa6ecf4d091ac12cR585. I think you should expose an method returning a bool for if lazy or not explicitly.

adityagoel4512: I see that it says it's unreachable in that case, but in principle it is user definable code and `ValueError` can emerge from anywhere so it makes little sense to assume it's because the array is lazy.

ndonnx/_typed_array/onnx.py
```diff
@@ -0,0 +1,2125 @@
...
+        axis_ = axis
+        if axis_ is None:
+            x = x.reshape((-1,))
+            axis_ = 0
+        res = TyArrayInt64(spox_op(x._var, axis=axis_, keepdims=keepdims))
+        if axis is None:
+            if keepdims:
+                res = res.reshape(tuple(1 for _ in range(self.ndim)))
+            # else:
+            #     res = res.squeeze(0)
```
adityagoel4512, line 938: ?

README.md
```diff
@@ -5,7 +5,7 @@
 [![conda-forge](https://img.shields.io/conda/vn/conda-forge/ndonnx?style=flat-square&logoColor=white&logo=conda-forge)](https://anaconda.org/conda-forge/ndonnx)
 [![pypi](https://img.shields.io/pypi/v/ndonnx.svg?logo=pypi&logoColor=white)](https://pypi.org/project/ndonnx)
 
-An ONNX-backed array library that is compliant with the [Array API](https://data-apis.org/array-api/) standard.
+An ONNX-backed implementation of the [Array API](https://data-apis.org/array-api/) standard.
```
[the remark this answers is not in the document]
cbourjau: Good point! I reduced the changes in the README to the minimum.

ndonnx/__init__.py
```diff
@@ -124,218 +176,197 @@
...
+    "Floating",
     "Integral",
+    "Numerical",
+    "as_nullable",
     "CoreType",
-    "CastError",
-    "promote_nullable",
-    "from_numpy_dtype",
     "UnsupportedOperationError",
+    "_TyArray",
```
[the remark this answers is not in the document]
cbourjau: It is no longer exported.

ndonnx/__init__.py
```diff
@@ -124,218 +176,197 @@
...
-    "Nullable",
-    "NullableCore",
+    "Floating",
     "Integral",
+    "Numerical",
+    "as_nullable",
     "CoreType",
-    "CastError",
-    "promote_nullable",
-    "from_numpy_dtype",
```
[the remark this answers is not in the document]
cbourjau: `from_numpy_dtype` is back in the game and exported publicly.

ndonnx/_array.py
```diff
@@ -1,614 +1,443 @@
...
-            and index.dtype != dtypes.bool
+        if not (
+            (True, True, False) == (is_shape, is_dtype, is_value)
+            or (False, False, True) == (is_shape, is_dtype, is_value)
+            or (False, False, False) == (is_shape, is_dtype, is_value)
         ):
-            raise TypeError(
-                f"Index must be an integral or boolean 'Array', not `{index.dtype}`"
-            )
+            raise ValueError("Invalid arguments.")
```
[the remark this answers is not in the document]
cbourjau: I reverted to using `ndonnx.asarray` and `ndonnx.array` (for arguments). I still find the latter terrible, but let's find something better in a follow-up. Thus, this constructor is no more.

ndonnx/_array.py
```diff
@@ -1,614 +1,443 @@
...
 
-        updates = ndx.asarray(updates).astype(self.dtype)
+        if isinstance(shape, tuple) and isinstance(dtype, DType):
+            self._tyarray = dtype.__ndx_argument__(shape)
+            return
+        if isinstance(value, np.ndarray):
+            raise NotImplementedError
```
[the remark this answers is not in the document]
cbourjau: I removed the constructor.

ndonnx/_typed_array/funcs.py
cbourjau: I straightened out the dependencies. It isn't perfect now, but much better, IMHO.

ndonnx/_typed_array/categorical.py
```diff
@@ -0,0 +1,244 @@
+# Copyright (c) QuantCo 2023-2025
+# SPDX-License-Identifier: BSD-3-Clause
+"""Implementation of "custom" datetime-related data types."""
```
adityagoel4512, line 3: Not datetime but categorical.

ndonnx/_typed_array/onnx.py
cbourjau: Well spotted!

ndonnx/_typed_array/onnx.py
cbourjau: I'm afraid the link is broken, but this is now done using the new `is_constant` property.

ndonnx/_array.py
```diff
@@ -1,614 +1,443 @@
...
-        """Number of elements in the array.
+    def __dlpack__(
+        self,
+        *,
+        stream: int | Any | None = None,
+        max_version: tuple[int, int] | None = None,
+        dl_device: tuple[Enum, int] | None = None,
+        copy: bool | None = None,
+    ) -> Any:
+        raise BufferError("ndonnx does not (yet) support the export of array data")
```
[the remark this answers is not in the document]
cbourjau: My thinking is that it can be done in the same way we support `unwarp_numpy` today: it would work for constant arrays, but not lazy ones.

ndonnx/_typed_array/onnx.py
```diff
@@ -0,0 +1,2101 @@
+# Copyright (c) QuantCo 2023-2025
```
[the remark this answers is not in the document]
cbourjau: I understand what you are saying and would be happy to find a different name, but it should be succinct. As it stands right now there are many places where the module is used (e.g. `onnx.int64` or `onnx.where`), and it would be quite unfortunate to always have to type `primitive.int64`. I like using the modules since it nicely scopes where things are coming from.

ndonnx/_dtypes.py
```diff
@@ -0,0 +1,156 @@
...
+        return onnx.float64
+
+    if np_dtype == np.bool_:
+        return onnx.bool_
+
+    if np_dtype.kind == "M":
+        unit = np.datetime_data(np_dtype)[0]
+        # Narrowing strings to Literal is virtually unsupported by
+        # mypy. The constructor raises if the unit is malformed,
+        # though.
```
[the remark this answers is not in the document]
cbourjau: It is in typing-extensions, but I used a simple validation function instead.

ndonnx/_dtypes.py
```diff
@@ -0,0 +1,156 @@
...
+        return datetime.DateTime64DType(unit=unit)  # type: ignore
+
+    if np_dtype.kind == "m":
+        unit = np.datetime_data(np_dtype)[0]
+        return datetime.TimeDelta64DType(unit=unit)  # type: ignore
+
+    # "T" i.e. "Text" is the kind used for `StringDType` in numpy >= 2
+    # See https://numpy.org/neps/nep-0055-string_dtype.html#python-api-for-stringdtype
+    if np_dtype.kind in ["U", "T"]:
```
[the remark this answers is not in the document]
cbourjau: I removed support for kind `"T"` for now. I think this is a very interesting dtype to support, though!

ndonnx/_funcs.py
```diff
@@ -4,1011 +4,654 @@
...
+from ._namespace_info import Device
+from ._typed_array import funcs as tyfuncs
+from ._typed_array import onnx
 
 
-class UnsupportedOperationError(TypeError): ...
+def from_dlpack(
+    x: object, /, *, device: None = None, copy: bool | None = None
+) -> Array:
+    raise BufferError("ndonnx does not (yet) support the export of array data")
```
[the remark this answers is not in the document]
cbourjau: I don't see it. It might have been fixed along the way.

ndonnx/_funcs.py
```diff
@@ -4,1011 +4,654 @@
...
 
+def mean(
+    x: Array, /, *, axis: int | tuple[int, ...] | None = None, keepdims: bool = False
+) -> Array:
+    return Array._from_tyarray(x._tyarray.mean(axis=axis, keepdims=keepdims))
 
-def log10(x):
-    return _unary(x.dtype._ops.log10, x)
 
+def meshgrid(*arrays: Array, indexing: str = "xy") -> list[Array]:
```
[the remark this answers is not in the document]
cbourjau: You are correct, but the current signature is the one from the standard. Maybe we can open an upsteam issue to see if there is interest in constraining these arguments with `Literal` types.

ndonnx/_funcs.py
```diff
@@ -4,1011 +4,654 @@
...
-    while (x := next(it, None)) is not None:
-        ret = ret + numeric_like(x)
-    target_shape = shape(ret)
-    return [broadcast_to(x, target_shape) for x in arrays]
 
+def reshape(
+    x: Array, /, shape: tuple[int, ...] | Array, *, copy: bool | None = None
+) -> Array:
+    if copy is not None:
+        raise ValueError("'copy' semantics are not implemented, yet")
```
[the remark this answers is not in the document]
cbourjau: `copy` is not about the containing `Array` but whether the underlying memory is copied. It is possible to view the same buffer using different shapes:
```
>>> import numpy as np
>>> arr = np.arange(9)
>>> arr_view = np.reshape(arr, (3,3,), copy=False)
>>> arr.shape
(9,)
>>> arr_view.shape
(3, 3)
>>> arr_view[:] = 0
>>> arr
array([0, 0, 0, 0, 0, 0, 0, 0, 0])
```

I don't think we can support this edge case easily. That said, all is good if `copy=True`, so I did allow that now.

ndonnx/_funcs.py
```diff
@@ -4,1011 +4,654 @@
...
-def expand_dims(x, axis=0):
-    if (out := x.dtype._ops.expand_dims(x, axis)) is not NotImplemented:
-        return out
-    raise UnsupportedOperationError(
-        f"Unsupported operand type for expand_dims: '{x.dtype}'"
-    )
 
+def result_type(*arrays_and_dtypes: Array | DType) -> DType:
+    if len(arrays_and_dtypes) == 0:
+        ValueError("at least one array or dtype is required")
```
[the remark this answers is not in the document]
cbourjau: Python displays raised exceptions in the following way:
```python
raise ValueError("foo has happened")
>>> ValueError: foo has happend
```
So the lower case makes sense. Furthermore, it is common to not add punctuation to the end as is also done in the standard library:
```python
[][42]
>>> IndexError: list index out of range
```

I strive to format all exceptions in this way, even though there were some slip-ups in this PR that I have now corrected (i.e., start with lowercase and no trailing punctuation).

ndonnx/__init__.py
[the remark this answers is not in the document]
cbourjau: `object-dtype` is not part of this PR any longer.

ndonnx/_future.py
```diff
@@ -0,0 +1,65 @@
...
+
+F = TypeVar("F", bound=Callable[..., Any])
+
+
+def _run_onnxruntime(
+    model: onnx.ModelProto,
+    input_feed: dict[str, spox._value_prop.ORTValue],
+    shared_library: str | Path,
+) -> dict[str, spox._value_prop.ORTValue]:
+    import onnxruntime
```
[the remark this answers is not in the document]
cbourjau: We still propagate values using the reference implementation if onnxruntime is not installed. Constant folding is no longer done. Nonetheless, the point is taken. There is now a warning telling the user about what is going on.

ndonnx/_typed_array/masked_onnx.py
cbourjau: This now uses the same names as in `onnx.py` since the information is already in the module name.

ndonnx/_typed_array/masked_onnx.py
cbourjau: The problem you describe is only relevant for the outputs, not the inputs. If a user creates a graph input using `ndx.argument(("N",), ndx.nint64)`, then the model will always require the mask to be provided unless it was not used in the graph and the model was built by passing `drop_unused=True` to `ndx.build` explicitly. 

Concerning the outputs, I followed your advice: `disassemble` now returns a mask-array in all cases.

ndonnx/_typed_array/masked_onnx.py
cbourjau: We have to specify the array class somewhere where it is accessible at runtime (i.e. not the generic type variable). The `_build` function is usually the place to do so.

tests/bugs/test_upstream_bugs.py
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) QuantCo 2023-2024
+# SPDX-License-Identifier: BSD-3-Clause
+
+import numpy as np
```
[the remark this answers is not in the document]
cbourjau: These tests run in 0.3s. They track issues we ran into and will alert us once we can exploit a fix. While not ideal, it isn't worth building anything more elaborate IMO.

ndonnx/_typed_array/masked_onnx.py
adityagoel4512: One other solution would be to add state to the dtype itself (a flag) that indicates if the null field is there or not.

CHANGELOG.rst
```diff
@@ -7,16 +7,85 @@
...
+  - :func:`~ndonnx.extensions.is_nullable_integer_dtype`
+  - :func:`~ndonnx.extensions.is_numeric_dtype`
+  - :func:`~ndonnx.extensions.is_onnx_dtype`
+  - :func:`~ndonnx.extensions.is_signed_integer_dtype`
+  - :func:`~ndonnx.extensions.s_time_unit`
+  - :func:`~ndonnx.extensions.is_unsigned_integer_dtype`
+  - :func:`~ndonnx.extensions.isin`
+  - :func:`~ndonnx.extensions.make_nullable`
+  - :func:`~ndonnx.extensions.put`
+  - :func:`~ndonnx.extensions.shape` (deprecated in favor of :func:`ndonnx.Array.dynamic_shape`)
```
adityagoel4512, line 62: It seems odd to move the function and immediately deprecate it. User code will break regardless, so might as well just delete the function entirely?

ndonnx/__init__.py
```diff
@@ -1,35 +1,45 @@
...
+
+# Set onnxruntime as the default for value propagation if it is
+# installed. We do so before the other imports so that any constant
+# that may be computed elsewhere is already using onnxruntime.
+if importlib.util.find_spec("onnxruntime"):
+    spox._value_prop._VALUE_PROP_BACKEND = spox._value_prop.ValuePropBackend.ONNXRUNTIME
+else:
+    warn(
+        "onnxruntime is not install. "
+        "Ndonnx will use the (incomplete) reference implementation for value propagation."
```
adityagoel4512, line 18:
```suggestion
        "onnxruntime is not installed. "
        "ndonnx will use the (incomplete) reference implementation for value propagation."
```

ndonnx/_typed_array/indexing.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+def _get_indices(
+    s: slice, length: int | TyArrayInt64
+) -> tuple[TyArrayInt64, TyArrayInt64, TyArrayInt64]:
+    """Array-compatible implementation of ``slice.indices``.
+
+    Slice members must be of type ``int`` or ``None`` while `length` may be an array.
+    """
+    from .funcs import maximum, minimum
```
adityagoel4512, line 221: I believe this lazy import can go by either directly using the Vars (we know the dtype is i64) or by using the dunder members.

ndonnx/_typed_array/onnx.py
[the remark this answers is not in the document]
adityagoel4512: You could always do `import ndonnx...primitive as prim` or something in that vein?

ndonnx/extensions.py
```diff
@@ -0,0 +1,437 @@
...
+
+    Returns
+    -------
+    out: Array
+        Array of booleans indicating whether each element of ``x`` is in ``items``.
+    """
+    return ndx.Array._from_tyarray(x._tyarray.isin(items))
+
+
+# TODO: Bad naming: It is obvious from the type hints that this mapping is "static"
```
adityagoel4512, line 64:
```suggestion
```

I would dispute that it's a "bad" name to say the mapping is static, but in any case, we haven't been shy to break the API here so I would either remove the comment or change the name if you're keen on doing so.

ndonnx/extensions.py
```diff
@@ -0,0 +1,437 @@
...
+        result_type = ndx.result_type(xty.data.dtype, value_.dtype)
+        if xty.mask is None:
+            res: tydx.TyArrayBase = xty.data
+        else:
+            res = tydx.funcs.where(xty.mask, value_, xty.data)
+    elif isinstance(xty, tydx.onnx.TyArray):
+        result_type = ndx.result_type(xty.dtype, value_.dtype)
+        res = xty.astype(result_type)
+    else:
+        raise TypeError("'fill_null' is only implemented for built-in types")
```
adityagoel4512, line 140: I believe this is a regression (see https://github.com/Quantco/ndonnx/blob/main/ndonnx/additional/_additional.py#L115, any dtype should be able to implement this). Can you offload it to the typed array in the style of the other functions?

ndonnx/extensions.py
```diff
@@ -0,0 +1,437 @@
...
+            default = 0  # type: ignore
+        elif values.dtype.kind == "U":
+            default = "MISSING"  # type: ignore
+        elif values.dtype == bool:
+            # to be in line with the numerical 0
+            default = False  # type: ignore
+    if default is None:
+        raise TypeError(
+            f"failed to infer default value for mapping values of data type `{values.dtype}`"
+        )
```
adityagoel4512, line 106: Is there a reason this isn't happening in the typed array implementation?

ndonnx/extensions.py
```diff
@@ -0,0 +1,437 @@
...
+        return ndx.Array._from_tyarray(tyarr)
+    if isinstance(x._tyarray, tydx.datetime.TimeBaseArray):
+        # TODO: The semantics of this branch are very odd!
+        is_nat = x._tyarray.is_nat
+        merged = tydx.masked_onnx.merge_masks(is_nat, null._tyarray)
+        if merged is not None:
+            is_nat = merged
+        return ndx.Array._from_tyarray(x._tyarray.dtype._build(data=x._tyarray._data))
+
+    raise ndx.UnsupportedOperationError(
```
adityagoel4512, line 207: You have deprecated this error but are raising it here. Does this not appear in the test suite?

ndonnx/extensions.py
```diff
@@ -0,0 +1,437 @@
...
+        tyarr = tydx.masked_onnx.make_nullable(x._tyarray.data, mask)
+        return ndx.Array._from_tyarray(tyarr)
+    if isinstance(x._tyarray, tydx.datetime.TimeBaseArray):
+        # TODO: The semantics of this branch are very odd!
+        is_nat = x._tyarray.is_nat
+        merged = tydx.masked_onnx.merge_masks(is_nat, null._tyarray)
+        if merged is not None:
+            is_nat = merged
+        return ndx.Array._from_tyarray(x._tyarray.dtype._build(data=x._tyarray._data))
+
```
adityagoel4512, line 206: Referencing my other comments, in this module there are _a lot_  of `isinstance` checks rather than dispatching the work to the typed array. If I want to define a new data type, I would not be able to define `make_nullable` or `fill_null` behaviour. If it is not applicable to the type, it won't be implemented. Consider the Polars list which we won't vendor but will define outside this project: https://docs.pola.rs/user-guide/expressions/lists-and-arrays/. This isn't an abstract consideration but rather used at this time, so if the regression is intentional we should be explicit about documenting it and providing a pattern for users to write generic code against their user-defined data types. This might lead to a proliferation of user-side functions that wrap these and dispatch to `ndonnx.extensions` if the dtype is vendored in ndonnx, which is unfortunate in my opinion.

ndonnx/extensions.py
```diff
@@ -0,0 +1,437 @@
...
+    """Return ``True`` if ``dtype`` is a nullable integer (i.e. "masked") data type."""
+    if isinstance(dtype, tydx.masked_onnx.FloatDTypes):
+        return True
+    return False
+
+
+def is_time_unit(s: str, /) -> TypeIs[tydx.datetime.Unit]:
+    if s in get_args(tydx.datetime.Unit):
+        return True
+    return False
```
adityagoel4512, line 415:
```suggestion
    return s in get_args(tydx.datetime.Unit)
```

refactoring.md
```diff
@@ -0,0 +1,247 @@
+# Summary
```
adityagoel4512, line 1: Happy to keep this in the repo for posterity but can you move it into a folder.

tests/bugs/test_upstream_bugs.py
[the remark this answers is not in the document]
adityagoel4512: Yes but I think it's a little arbitrary to draw the line here for elaborate when we are doing property testing with the Array API in CI weekly already 😆.

tests/test_datetime.py
```diff
@@ -0,0 +1,340 @@
...
+def test_datetime_from_np_array(ty, unit):
+    np_arr = np.array([0, 1], f"{ty}[{unit}]")
+    arr = ndx.asarray(np_arr)
+    np.testing.assert_array_equal(arr.unwrap_numpy(), np_arr, strict=True)
+
+
+@pytest.mark.parametrize(
+    "scalar, dtype, res_dtype",
+    [
+        # TODO: Implement other units
```
adityagoel4512, line 59: We have more units now?

tests/test_datetime.py
```diff
@@ -0,0 +1,340 @@
...
+    assert_equal_dtype_shape(arr + scalar, res_dtype, shape)
+
+
+@pytest.mark.parametrize(
+    "op",
+    [
+        operator.add,
+        operator.sub,
+        operator.mul,
+        # operator.truediv
```
adityagoel4512, line 77:
```suggestion
        operator.truediv
```

ndonnx/extensions.py
cbourjau: The breaking change in the public API is actually quite limited.

refactoring.md
cbourjau: I'm happy to remove it before merging, too. It will remain available in this PR.

CHANGELOG.rst
adityagoel4512: I meant that there since we expect other intentional user-breaking changes in this release, it seemed like an opportunity to also immediately do so for `shape` rather than give a deprecation it in the new module.

ndonnx/extensions.py
cbourjau: There is no standardized notion of what `null` is meant to be for a general data type and I'd like to keep it that way. In these terms it does not make sense to have this function be implemented by custom data types.

ndonnx/extensions.py
cbourjau: There are many deprecation warnings raised by the tests suite at the moment. I'd rather fix them in a follow-up PR. Functionally, `ndx.UnsupportedOperationError` is an alias for `TypeError` now. Nonetheless, it shouldn't be used here.

ndonnx/extensions.py
cbourjau: As written in the other comment, there is no standardized notion what `null` is supposed to be and I don't think there should be. I am not convinced that nullable data types should be shipped with ndonnx in the future, so I'd like to keep them confined for now. I added a note about `make_nullable` not being accessible for custom data types.

ndonnx/extensions.py
cbourjau: I clarified the comment.

ndonnx/extensions.py
cbourjau: I added the defaults to the docstring now, which I think gives this code its logical place here. The typed array member functions do not have an optional `default`.

tests/test_datetime.py
cbourjau: I parametrized many of the tests over all units now.

tests/bugs/test_upstream_bugs.py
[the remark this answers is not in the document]
cbourjau: The weekly Array API CI tests run for almost 20 minutes. It's reasonable to invest a little effort to not have them run on every commit.

ndonnx/_typed_array/onnx.py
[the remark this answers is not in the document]
cbourjau: I also like the name `onnx` since it is very on-point: These are types with a direct representation in the ONNX standard. I'd like to leave it as-is unless you have a strong opinion on this.

docs/experimental/experimental.rst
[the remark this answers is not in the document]
cbourjau: After thinking about this a bit more, I don't think it is worth the effort to document this in-depth at this point. I removed the item entirely for now.

refactoring.md
cbourjau: Actually, I just moved it into `/docs`. Couldn't get myself to delete it :D

docs/experimental/experimental.rst
```diff
@@ -1,160 +1,27 @@
...
-
-    @eager_propagate
+    @propagate_with_custom_operators("path/to/shared_library.so")
     def my_op(x: ndx.Array) -> ndx.Array:
-        return ndx.from_spox_var(my_op(x.spox_var()))
+        return ndx.asarray(my_op(x.disassemble()))
 
 .. warning::
     Using custom operators may limit the portability of your exported ONNX model since not all runtimes may implement a non-standard operator.
-
```
adityagoel4512, line 53: We agreed to revisit this in the future.

docs/refactoring.md
```diff
@@ -0,0 +1,247 @@
+# Summary
```
adityagoel4512, line 1: Remove? Might be nice to attach it to the PR description for posterity.

tests/bugs/test_upstream_bugs.py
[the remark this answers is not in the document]
adityagoel4512: I cannot say I agree with running upstream bug checks in the unit tests but I'd like to move the rest of this PR along so we can revisit this in the future.

tests/test_masked.py
```diff
@@ -173,6 +152,7 @@ def test_asarray_masked():
     )
 
 
+@pytest.mark.skip(reason="no longer relevant")
```
adityagoel4512, line 155: Just remove the test?

ndonnx/_typed_array/categorical.py
```diff
@@ -0,0 +1,244 @@
...
+        GetitemIndex,
+        SetitemIndex,
+    )
+    from .onnx import TyArrayInt64
+
+
+_N_MAX_CATEGORIES = np.iinfo(np.uint16).max
+
+
+class CategoricalDType(DType["CategoricalArray"]):
```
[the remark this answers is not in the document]
adityagoel4512: Before merging, please remove categorical from the PR.

adityagoel4512 (approved):
Thanks for the substantial effort. I've left a couple of comments prior to merging but otherwise let's give this a go 🚀. We can revisit a couple of items in a future PR.

ndonnx/extensions.py
```diff
@@ -0,0 +1,437 @@
...
+def put(a: ndx.Array, indices: ndx.Array, updates: ndx.Array, /) -> None:
+    """Replaces specified elements of an array with given values.
+
+    This function follows the semantics of `numpy.put` with
+    `mode="raises". The data types of the update array and the updates
+    must match. The indices must be provided as a 1D int64 array.
+    """
+    from ._typed_array.onnx import TyArrayInt64
+
+    if not isinstance(indices._tyarray, TyArrayInt64):
```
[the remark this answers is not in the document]
cbourjau: `isinstance` is needed to get the type narrowing of `indices._tyarray` further down. I added a comment.

docs/refactoring.md
cbourjau: I thought it could stay in the docs folder. Or did I read one of your earlier comments on this wrong?

adityagoel4512: I don't mind it staying around for a short while. I thought you yourself said you'd remove it.

cbourjau: I did, but then changed my mind :D

PR #111, opened 2025-03-13
Add tests for schema stability

cbourjau:
This test is back-ported from #110. It should be its own PR based on the current `main` since it ought to ensure that we did not break the current schemas.

### DIFF
--- a/tests/test_build_utils.py
+++ b/tests/test_build_utils.py
@@ -1,4 +1,4 @@
-# Copyright (c) QuantCo 2023-2024
+# Copyright (c) QuantCo 2023-2025
 # SPDX-License-Identifier: BSD-3-Clause
 
 import json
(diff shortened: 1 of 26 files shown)


tests/test_build_utils.py
```diff
@@ -38,3 +42,112 @@ def test_input_output_name_backwards_compatibility(dtype):
...
+    else:
+        dtype = str(dtype).lower()
+    fname = Path(__file__).parent / f"schemas/{dtype}.json"
+
+    # Only set to `True` temporarily if the schemas need to be
+    # re-generated. A schema, once stabilized, must never
+    # change. However, this flag may come in handy if we add new data
+    # types for instance.
+    update = False
+    if update:
```
adityagoel4512, line 136: I'd remove the update flag from the test suite and have this in an independent script. You can make a pixi task for generating schemas from the dtype submodule if you'd like some automation around this.

cbourjau: I like having the code here since it directly hooks into the parametrization of the test. We would need to maintain a separate list of data types for the script along with the path etc. Also, running a pixi task is probably easier than we want the generation to be. I'd prefer to keep things like this.

adityagoel4512: A pytest argument will do; just no source code manipulation to generate schemas please!

adityagoel4512 (approved):

PR #121, opened 2025-04-30
Support the 2024.12 version of the Array API

cbourjau:
Support most of the newly added functions of the 2024.12 standard. Unfortunately, I don't see a way of supporting `nextafter` without a new operator in the ONNX standard. The `reduce_prod` function is supported using `log` and `exp`, which are quite expensive. We should push for an addition to ONNX standard in this regard, too. 

The newly added integer array indexing prompted a minor refactor of the current indexing logic.


ndonnx/_typed_array/_utils.py
```diff
@@ -57,12 +57,18 @@ def safe_cast(ty: type[T], a: TyArrayBase | bool) -> T:
...
+    x2: TyArrayBase | PyScalar,
+    result: T | NotImplementedType,
+    func_name: str,
 ) -> T:
-    """Validate if the provided result is not `NotImplemented` and raise an error if it
-    is."""
+    """Raise an exception if `result` is not `NotImplemented`, otherwise return it."""
+
+    def fmt(x: TyArrayBase | PyScalar) -> str:
+        return str(x) if isinstance(x, PyScalar) else str(x.dtype)
```
adityagoel4512, line 68:
```suggestion
        return type(x) if isinstance(x, PyScalar) else str(x.dtype)
```

Seems a little strange to produce a dtype for the array and the value for the scalar.

ndonnx/_typed_array/funcs.py
```diff
@@ -193,34 +194,61 @@ def where(cond: onnx.TyArrayBool, x: TYARRAY, y: TYARRAY) -> TYARRAY: ...
...
+
+    if isinstance(x1, TyArrayBase):
+        res = x1.__ndx_logaddexp__(x2)
+    if res is NotImplemented and isinstance(x2, TyArrayBase):
+        res = x2.__ndx_logaddexp__(x1)
+    return validate_op_result(x1, x2, res, "logaddexp")
+
+
+def logical_and(x1: TyArrayBase | bool, x2: TyArrayBase | bool, /) -> TyArrayBase:
+    x1 = onnx.const(x1) if isinstance(x1, bool) else x1
```
adityagoel4512, line 229: Why is this not just `x1 & x2`? I would have expected the dunder methods will take care of producing a constant where required and ensuring things are dispatched onto the the correct typed array implementation.

cbourjau: `x1 & x2` is `bitwise_and`. The difference is minimal, but the latter is guaranteed to return a boolean data type while the former is not. Either way, we want to dispatch to the correct underlying method and that is easier if we only deal with arrays.

adityagoel4512: Ok then I would suggest we should be consistent. For instance, `ndx.add` for scalars does a python addition and then wraps the result in an array. This has different semantics to passing two 0D ndonnx arrays through (consider overflows and the fact that python integers are unbounded in size). 

Clearly the first case below overflows, but I don't see any special treatment for python scalars in the standard and I think it is non-intuitive that a python scalar and a 0D array behave differently in an ndonnx function:

```python
>>> ndx.add(ndx.asarray(1<<63 -1), ndx.asarray(1<<63 -1))
array(data: -9223372036854775808, dtype=int64)
>>> ndx.add(1<<63 -1, 1<<63 -1)
Traceback (most recent call last):
  File "<python-input-15>", line 1, in <module>
    ndx.add(1<<63 -1, 1<<63 -1)
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/Users/adi/Documents/ndonnx/ndonnx/_elementwise.py", line 18, in add
    return ndx.asarray(a + b)
           ~~~~~~~~~~~^^^^^^^
  File "/Users/adi/Documents/ndonnx/ndonnx/_funcs.py", line 85, in asarray
    return Array._from_tyarray(tyfuncs.astyarray(obj, dtype=dtype))
                               ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/Users/adi/Documents/ndonnx/ndonnx/_typed_array/funcs.py", line 95, in astyarray
    res = inferred_dtype.__ndx_create__(val)
  File "/Users/adi/Documents/ndonnx/ndonnx/_typed_array/onnx.py", line 117, in __ndx_create__
    return const(val, dtype=self)
  File "/Users/adi/Documents/ndonnx/ndonnx/_typed_array/onnx.py", line 2156, in const
    obj = np.asarray(obj, dtype=np.int64)
OverflowError: Python int too large to convert to C long
```

adityagoel4512: Okay, I don't love the idea of deviating from NumPy unnecessarily but let's get this merged and review the overflow semantics we would like in the follow up.

adityagoel4512 (approved):

ndonnx/_typed_array/funcs.py
adityagoel4512: The other issue I thought of is that for user-defined data types, addition may now need to be defined against `int` and `float` as well as arrays. This is another argument for first constructing an array.

cbourjau: Sorry, my comment was maybe ambiguous. I meant to say that requiring at least one argument to be an array is what we should do here. This means that we don't deviate from NumPy but rather say within the subset of its semantics. Are we on the same page?

adityagoel4512: I'm fine with being a subset! We're now here, which is a reasonable outcome to me:

```python
>>> import numpy as np
>>> import ndonnx as ndx
>>> np.add(1, 2)
np.int64(3)
>>> ndx.add(1, 2)
Traceback (most recent call last):
  File "<python-input-3>", line 1, in <module>
    ndx.add(1, 2)
    ~~~~~~~^^^^^^
  File "/Users/adi/Documents/ndonnx/ndonnx/_elementwise.py", line 29, in wrapped
    raise TypeError(
        f"at least one argument to '{fn.__name__}' must be of type 'ndonnx.Array'"
    )
TypeError: at least one argument to 'add' must be of type 'ndonnx.Array'
```

PR #126, opened 2025-05-12
Add workflow to compute OpenSSF scorecard

xhochy:
To better assess the quality/maintenance of an open-source project the OpenSSF scorecard is a standard tool. We would like to compute and provide the values for all our projects. You can see an example at https://scorecard.dev/viewer/?uri=github.com/Quantco/pixi-pack

We will also be able to see the results in the Security tab.


adityagoel4512 (approved):
Thanks Uwe

PR #127, opened 2025-05-12
Fix modulus

adityagoel4512:
Closes https://github.com/Quantco/ndonnx/issues/125. ORT arrives at some UB when doing modulus of zero: https://github.com/microsoft/onnxruntime/blob/7d719758f34ef99123cc26aa18bbfda508108c57/onnxruntime/core/providers/cpu/math/element_wise_ops.cc#L2010C14-L2010C20.


cbourjau (approved):
Thanks for debugging and fixing this!

PR #135, opened 2025-05-23
Mitigate onnxruntime segault in matmul due to zero-sized inputs

cbourjau:
This one was a chore! In the past, we already ran into segfaults/errors from onnxruntime with matmul. However, the mitigation that we [had in place](https://github.com/Quantco/ndonnx/pull/102/files) was slightly unsound and didn't survive the recent refactoring. We used an `If` node with a matmul-branch and another where we returned a zero-sized dummy. However, the zero-sized dummy did not have the correct shape/rank. The solution proposed here always returns the correct shape and rank. While it looks very involved, it is not too bad computationally compared to the matmul that is about to take place.

### DIFF
--- a/ndonnx/_typed_array/onnx.py
+++ b/ndonnx/_typed_array/onnx.py
@@ -10,7 +10,7 @@ from types import EllipsisType, NotImplementedType
 from typing import Literal, TypeAlias, TypeGuard, TypeVar, cast, overload
 
 import numpy as np
-from spox import Tensor, Var, argument
+from spox import Tensor, Var, argument, build, inline
 from typing_extensions import Self
 
 from ndonnx import DType
@@ -19,7 +19,7 @@ from ndonnx.types import NestedSequence, OnnxShape, PyScalar
 from .._schema import DTypeInfoV1
 from . import TyArrayBase, safe_cast
 from . import ort_compat as op
-from .dtype_independent_funcs import maximum, minimum, where
+from .dtype_independent_funcs import maximum, minimum, where, zeros
 from .indexing import FancySlice
 
 _ScalarInt: TypeAlias = "TyArrayInteger"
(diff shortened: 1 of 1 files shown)


ndonnx/_typed_array/onnx.py
```diff
@@ -2633,3 +2639,86 @@ def numeric_like(x: TyArrayBase) -> TyArrayInt32:
...
+
+    (var,) = op.if_(
+        safe_cast(TyArrayBool, (a_.dynamic_size * b_.dynamic_size == 0))._var,
+        then_branch=lambda: [dummy_out],
+        else_branch=lambda: [res],
+    )
+    model = build({"a": a_._var, "b": b_._var}, {"res": var})
+    (var_with_propagation,) = inline(model)(a=a._var, b=b._var).values()
+
+    arr = type(a)(var_with_propagation)
```
adityagoel4512, line 2723: This section seems more appropriate for `ort_compat` for guarding against zero dimensions. I get why the previous bit would stay here.

cbourjau: But the previous part is needed to compute `dummy_out`. I'm not quite seeing how we would be able to nicely move this part into `ort_compat.py`

adityagoel4512: I was imaging providing dummy and result branches as zero arg lambdas and then doing something like `opx.guard_zero_dims(evaluate=lambda: ..., fallback=lambda:...)`.

cbourjau: Thus far everything in `ort_compat` just mirrors standard operators. Ideally, we would have a `ort_compat.matmul` that Does The Right Thing. Then, once we have things patched upstream we can just remove the workaround in `ort_compat` and profit everywhere without touching any code elsewhere. However, I'm not sure how we would use the proposed `guard_zero_dims` in this context. What we could do, and what may actually be nicer, is that we do move this function into `ort_compat` but use lazy imports to make it work. What do you think of that?

PR #138, opened 2025-05-25
Update array api tests and fixes

cbourjau:
This updates the array-api-tests submodule and introduces some fixes. It further cranks up the number of examples we run on our weekly CI.

The array-api-test suite got more efficient to execute recently (the regular CI job finishes in less than a minute). I want to update the number of examples we run on our regular CI too, but this quickly surfaces #94 (see [here](https://github.com/Quantco/ndonnx/actions/runs/15239141231/job/42856990826#step:6:360)). We could increase the examples now and skip the relevant tests entirely, but I think it is better to at least test some examples.

### DIFF
--- a/ndonnx/__init__.py
+++ b/ndonnx/__init__.py
@@ -210,6 +210,7 @@ _default_float = float64
 __array_api_version__ = "2024.12"
 
 __all__ = [
+    "__array_api_version__",
     "__array_namespace_info__",
     "DType",
     # ONNX data types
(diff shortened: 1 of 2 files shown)


adityagoel4512 (approved):

ndonnx/__init__.py
```diff
@@ -207,6 +207,8 @@
 _default_int = int64
 _default_float = float64
 
+__array_api_version__ = "2024.12"
+
 __all__ = [
     "__array_namespace_info__",
```
adityagoel4512, line 213:
```suggestion
    "__array_namespace_info__",
    "__array_api_version__",
```

PR #149, opened 2025-07-02
Fix rank dynamic size

cbourjau:
Calling `Array.dynamic_size` should return a scalar in all cases. However, the current release of ndonnx returns a 1D array (with a single element) if `dynamic_size` is called on 1D input.

### DIFF
--- a/ndonnx/_array.py
+++ b/ndonnx/_array.py
@@ -120,7 +120,7 @@ class Array:
         if self.ndim == 0:
             Array._from_tyarray(onnx.const(1, dtype=onnx.int64))
         if self.ndim == 1:
-            return self.dynamic_shape
+            return self.dynamic_shape[0]
 
         return Array._from_tyarray(self.dynamic_shape._tyarray.prod())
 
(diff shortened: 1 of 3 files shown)


ndonnx/_array.py
```diff
@@ -120,7 +120,7 @@ def dynamic_size(self) -> Array:
         if self.ndim == 0:
             Array._from_tyarray(onnx.const(1, dtype=onnx.int64))
         if self.ndim == 1:
-            return self.dynamic_shape
+            return self.dynamic_shape[0]
 
         return Array._from_tyarray(self.dynamic_shape._tyarray.prod())
```
adityagoel4512, line 125: We have a scalar type right?

cbourjau: I'm not sure I understand what you mean. Indexing into an array-api-array always returns an array, but that array may be a rank-0 one (as is the case here). We don't have an explicit type for that though.

adityagoel4512: That's right, I briefly mistook the utility of PyScalar.

PR #156, opened 2025-07-17
Improve support for NumPy generics

cbourjau:
ndonnx arrays already somewhat work with `numpy.generics` in some circumstances. The current support is rather by accident, but it is a direction we wanted to go towards in any case.

This PR fixes a type promotion issue in which NumPy generics were wrongly handled as regular Python scalars. Furthermore, this it fixes an issue in which time-generics led to an erroneous exception.

### DIFF
--- a/ndonnx/_typed_array/funcs.py
+++ b/ndonnx/_typed_array/funcs.py
@@ -36,6 +36,8 @@ def _infer_sequence(
 def _infer_dtype(
     val: PyScalar | np.ndarray | TyArrayBase | Var | NestedSequence | np.generic,
 ) -> DType:
+    if isinstance(val, np.generic):
+        return from_numpy_dtype(val.dtype)
     if isinstance(val, np.ndarray):
         if val.dtype == object:
             if not all(isinstance(el, str) for el in val.flatten()):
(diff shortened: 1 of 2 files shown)


ndonnx/_typed_array/funcs.py
```diff
@@ -64,9 +64,10 @@ def _infer_dtype(
     elif isinstance(val, Sequence):
         return _infer_sequence(val)
     elif isinstance(val, np.generic):
+        if val.dtype.kind in "Mm":
+            return datetime.from_numpy_dtype(val.dtype)
         return onnx.from_numpy_dtype(val.dtype)
```
adityagoel4512, line 69: Can we consolidate this into one function for determining the ndonnx data type from the numpy dtype?

adityagoel4512, line 69: Can't we just use `from_numpy_dtype` here?

ndonnx/types.py
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) QuantCo 2023-2025
 # SPDX-License-Identifier: BSD-3-Clause
 
+from __future__ import annotations
+
```
adityagoel4512, line 5:
```suggestion
```

PR #157, opened 2025-07-17
Correctly return NotImplemented from dunder methods for unsupported operands 

cbourjau:
The dunder methods of `ndonnx.Array` should return `NotImplemented` when faced with an unsupported type, but they don't currently. This PR fixes this oversight.


adityagoel4512 (changes requested):

ndonnx/_array.py
```diff
@@ -23,18 +23,86 @@
...
+    #   -> Option 1 (Taken in the current implementation):
+    #     -> operator.add(TyArray(dtype1), TyArray(dtype2)):
+    #        - Tries __add__ and __radd__ on the TyArray objects
+    #        - Raise TypeError with bad error message
+    #        - Catch and raise new error with better message
+    #   -> Option 2:
+    #     -> Manually try __add__ and __radd__ on the TyArray objects
+    #       -> Pass through the returned NotImplemented object
+    #       -> Still a bad error message for the user, but does not expose internal class names
+    #   -> Option 3:
```
adityagoel4512, line 65: Can you please just check for `NotImplemented` on both implementations (i.e. option 3)? This try ... except logic feels brittle as all TypeErrors are now being trapped here and rewritten, which forces all the typed arrays to be extra careful when raising TypeErrors as they are all interpreted as being due to operand dtype mismatch. I'm especially concerned about custom dtypes as this behavior is undocumented.

tests/test_interaction_numpy.py
```diff
@@ -43,15 +43,18 @@
     ],
 )
 def test_dunders_numpy_generic(op, np_arr, np_gen):
-    # Forwards
+    # The first operand is multiplied by two to better test the
+    # correct application of non-cummutative functions.
```
adityagoel4512, line 47:
```suggestion
    # correct application of non-commutative functions.
```

adityagoel4512 (approved):

PR #161, opened 2025-08-12
(Re)implement ndonnx.repeat for all built-in data types

cbourjau:
This PR implements `repeat` for all built-in data types. This is possible since the implementation only relies on other high-level APIs instead of the underlying layout. We used to have an implementation for `ndonnx.repeat` before, but it got lost during the refactoring leading up to 0.10.0. The previous implementation used `searchsorted` while this one uses broadcasting and reshaping. Either has compromises. The earlier one was likely heavier on the CPU, while this one may be more memory hungry in some circumstances. My gut feeling is that this implementation is the better compromise.

The change to the array-api-tests is due to this upstream issue: https://github.com/data-apis/array-api-tests/pull/388

@justinchuby


ndonnx/_typed_array/typed_array.py
```diff
@@ -457,7 +457,58 @@ def variance(
...
+            key[axis + 1] = None
+            tmp = x[tuple(key)].broadcast_to(
+                x_shape[: axis + 1].concat([repeats, x_shape[axis + 1 :]], axis=0)
+            )
+            # We now have a shape of (..., self.shape[axis], repeats, ...).
+            # All we need to do is collapse these two axes into one.
+            out_shape = x_shape.copy()
+            out_shape[axis] = safe_cast(onnx.TyArrayInt64, out_shape[axis] * repeats)
+
+            return tmp.reshape(out_shape)
```
adityagoel4512, line 494:
```suggestion
            x_shape[axis] = x_shape[axis] * repeats
            return tmp.reshape(x_shape)
```

It'd be better if you updated the type hint in the typed integer array class for multiplication rather than do this cast.

cbourjau: Fair enough. Let's do that next time!

PR #165, opened 2025-08-20
Allow arrays as inputs to ndonnx.arange

cbourjau:
Possibly the most requested feature is finally here 🥳 !

This PR does what it says in the title. Array inputs are essentially passed down to the individual implementations untouched, which may return `NotImplemented`. If no `dtype` is specified but at least one argument is an `Array`, we try the array dtypes until we find one that does not return `NotImplemented`.

### DIFF
--- a/ndonnx/_funcs.py
+++ b/ndonnx/_funcs.py
@@ -105,34 +105,35 @@ def any(
 
 
 def arange(
-    start: int | float,
+    start: int | float | Array,
     /,
-    stop: int | float | None = None,
-    step: int | float = 1,
+    stop: int | float | Array | None = None,
+    step: int | float | Array = 1,
     *,
     dtype: DType | None = None,
     device: None | Device = None,
 ) -> Array:
-    if dtype is None:
-        if builtins.all(
-            isinstance(el, int) for el in [start, stop, step] if el is not None
-        ):
-            dtype = ndx._default_int
-        else:
-            dtype = ndx._default_float
-    dtype = dtype or ndx._default_float
+    sss = [start, stop, step]
+    for item in sss:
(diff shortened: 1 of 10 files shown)


ndonnx/_funcs.py
```diff
@@ -105,34 +105,35 @@ def any(
...
         stop = start
         start = 0
 
-    for el in [start, stop, step]:
-        if not isinstance(el, int | float):
-            raise TypeError(
-                f"unexpected type for 'start', 'stop', or 'step': `{type(el)}`"
-            )
+    array_or_scalar = [
+        el._tyarray if isinstance(el, Array) else el for el in [start, stop, step]
```
adityagoel4512, line 133:
```suggestion
        el._tyarray if isinstance(el, Array) else el for el in sss
```

adityagoel4512 (approved):

ndonnx/_funcs.py
```diff
@@ -105,34 +105,35 @@ def any(
...
-    if dtype is None:
-        if builtins.all(
-            isinstance(el, int) for el in [start, stop, step] if el is not None
-        ):
-            dtype = ndx._default_int
-        else:
-            dtype = ndx._default_float
-    dtype = dtype or ndx._default_float
+    sss = [start, stop, step]
+    for item in sss:
```
cbourjau, line 117:
```suggestion
    for item in [start, stop, step]:
```

ndonnx/_funcs.py
cbourjau: `start`, `stop`, `step` were mutated in between. I removed `sss` since it provided no benefit here.

PR #167, opened 2025-08-25
Fix ndonnx.clip for integer dtype array with float type min/max values

konstantin-pueckler-qc:
``ndonnx.clip`` did not work correctly when given an integer dtype ``ndonnx.Array`` and float type values as min/max arguments because it did not cast the output array's type correctly.
I came across this bug when testing ndonnx/pdonnx support in quantcore.learn.

### DIFF
--- a/tests/test_free_functions.py
+++ b/tests/test_free_functions.py
@@ -189,7 +189,7 @@ def test_clip():
 )
 def test_int_float_clip():
     def do(npx):
-        return npx.clip(npx.asarray([1,2]), min=1.5, max=1.5)
+        return npx.clip(npx.asarray([1, 2]), min=1.5, max=1.5)
 
     np.testing.assert_array_equal(do(ndx).unwrap_numpy(), do(np))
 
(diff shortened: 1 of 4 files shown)


tests/test_free_functions.py
```diff
@@ -184,6 +184,16 @@ def do(npx):
     np.testing.assert_array_equal(do(ndx).unwrap_numpy(), do(np))
 
 
+@pytest.mark.skipif(
+    np.__version__ < "2", reason="'clip' has a different API in NumPy 1.x"
```
adityagoel4512, line 188: Different API or different semantics?

adityagoel4512:
We're following the Array API standard which you might have encountered: https://data-apis.org/array-api/2023.12/API_specification/generated/array_api.clip.html.  It looks like behavior is implementation-defined when the bounds' dtype differs to the array's, but aligning with NumPy and promoting between min, max and the array seems okay to me. Can you reframe the change log to explain what is changing along these lines?

tests/test_free_functions.py
konstantin-pueckler-qc: the only difference seems to be that in numpy 1.x the min and max arguments are positional only.
I made the arguments given to the call positional, now it works fine

adityagoel4512 (approved):

PR #169, opened 2025-08-28
fixed timedelta creation

konstantin-pueckler-qc:
currently, performing something like 
``arr = ndx.asarray(np.asarray([1]), dtype=ndx.TimeDelta64DType("ns"))``
``ndx.asarray(arr)`` 
raises an Exception, which is probably not what is wanted. I changed the ``__ndx_create__`` method of ``TimeDelta64DType`` so that now also ``TyArrayTimeDelta`` objects can be handled.
I also changed the ``__ndx_create__`` method of ``TyArrayDatetime64``. So far, if it received a ``TyArrayDatetime64`` , it only performed copy, ignoring the unit.

### DIFF
--- a/tests/test_creation.py
+++ b/tests/test_creation.py
@@ -72,6 +72,25 @@ def test_arange_array_arg(start, stop, step):
     np.testing.assert_array_equal(np_res[0], ndx_res[0], strict=True)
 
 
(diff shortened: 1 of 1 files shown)


cbourjau (approved):
Thanks!

PR #205, opened 2026-04-02
Implement `take` and `take_along_axis` for time dtypes

cbourjau:


SamuelLessmannQC (commented):
Is there a specific reason not to reshape like for `repeat` as done [here](https://github.com/Quantco/ndonnx/blob/a3ec7031ff4c27ec5527959d5c465e6f913b26ef/ndonnx/_typed_array/typed_array.py#L470)?

This fails:
```py
@pytest.mark.parametrize("dtype", ["datetime64[s]", "timedelta64[s]"])
def test_take_axis_none_flattens(dtype):
    np_arr = np.asarray(
        [
            [1, "NaT", 0],
            [10, 20, 30],
        ],
        dtype=dtype,
    )

    def do(npx):
        indices = np.asarray([4, 1, -1], dtype=np.int64)
        return npx.take(npx.asarray(np_arr), npx.asarray(indices))

    np.testing.assert_array_equal(do(ndx).unwrap_numpy(), do(np))
```

SamuelLessmannQC (approved):

PR #217, opened 2026-06-04
Drop support for numpy<2

cbourjau:


sgilm (approved):
Just a few stray comments to potentially get rid of:

```bash
> git grep -E "np1x|1\\.x"
ndonnx/_array.py:43:            # Note: NumPy generic are subclasses of Python scalars in np1x
ndonnx/_array.py:69:            # Note: NumPy generic are subclasses of Python scalars in np1x
ndonnx/_typed_array/onnx.py:988:        # Compat for Windows on numpy 1.x
ndonnx/_typed_array/onnx.py:2331:    # don't blindly fall back to NumPy to maintain better np1x
tests/test_interaction_numpy.py:48:    # NumPy 1.x converts np.uint32(2) * 2 to np.float64. Adding np_gen
```

Issue #131, opened 2025-05-21
Support type promotion between datetimes of different resolution


adityagoel4512:
NumPy permits type promotion inside `np.where` between datetimes of differing resolution. We should support this in ndonnx:

```bash
(Pdb) x
array(['1999-01-01T00:00:00', '2004-03-03T00:00:00'],
      dtype='datetime64[s]')
(Pdb) y
array(['2003-03-03T00:00:00.000000000', '2014-01-01T00:00:00.000000000'],
      dtype='datetime64[ns]')
(Pdb) cond
array([False,  True])
(Pdb) np.where(cond, x, y)
array(['2003-03-03T00:00:00.000000000', '2004-03-03T00:00:00.000000000'],
      dtype='datetime64[ns]')
(Pdb) ndx.where(ndx.asarray(cond), ndx.asarray(x), ndx.asarray(y))
*** TypeError: unsupported operand data types for 'where': `datetime64[s]` and `datetime64[ns]`
```

PR #218, opened 2026-06-10
Add optional prefix to ndonnx.build

EmilyMohellenbrandQC:
Currently, ndonnx.build() throws an error if the same field name exists in both inputs and outputs due to name collisions in the underlying ONNX graph.

To resolve this issue, this PR introduces optional `input_prefix` and `output_prefix` arguments to ndonnx.build(). This allows users to supply custom prefixes to distinguish input and output fields in the ONNX graph while retaining the ability to map them back to their original public names at inference time via schema metadata.

**Summary of Changes**

- ndonnx/ndonnx/_build.py:
   - Added optional `input_prefix` and `output_prefix` keyword-only arguments to build() (defaulting to "" for backwards compatibility).
   - Prepended prefixes to variable names inside build() when disassembling inputs/outputs to Spox variables.
   - Adjusted the `kept_inputs` check and the metadata `input_schema` and `output_schema` keys to correctly use **prefixed variable names**.

- ndonnx/ndonnx/_schema.py:
   - Updated the `SchemaV1` schema dataclass and its `parse_json()` method to store and parse `input_prefix` and `output_prefix`.
   - Updated `to_json()` to omit empty ("") and None prefixes to maintain strict backward compatibility with existing snapshot tests.

- ndonnx/tests/test_build_utils.py:
   - Added unit test `test_build_with_prefixes` to verify prefix prepending, schema metadata generation, and standard onnxruntime execution of prefixed models.

### DIFF
--- a/ndonnx/_build.py
+++ b/ndonnx/_build.py
@@ -37,8 +37,8 @@ def build(
     onnx.ModelProto
         ONNX model
     """
-    ins = _arrays_to_vars(inputs, prefix=input_prefix)
-    outs = _arrays_to_vars(outputs, prefix=output_prefix)
+    ins = {f"{input_prefix}{k}": v for k, v in _arrays_to_vars(inputs).items()}
+    outs = {f"{output_prefix}{k}": v for k, v in _arrays_to_vars(outputs).items()}
 
     mp = spox_build(ins, outs, drop_unused_inputs=drop_unused)
 
@@ -74,10 +74,10 @@ def build(
     return mp
 
 
-def _arrays_to_vars(dct_of_arrs: dict[str, Array], prefix: str = "") -> dict[str, Var]:
+def _arrays_to_vars(dct_of_arrs: dict[str, Array]) -> dict[str, Var]:
     out: dict[str, Var] = {}
     for name, arr in dct_of_arrs.items():
-        out |= _disassemble_named_array(f"{prefix}{name}", arr)
+        out |= _disassemble_named_array(name, arr)
     return out
 
 
--- a/ndonnx/_build.py
+++ b/ndonnx/_build.py
@@ -10,7 +10,12 @@ from ._schema import SchemaV1
 
 
 def build(
-    inputs: dict[str, Array], outputs: dict[str, Array], drop_unused=False
+    inputs: dict[str, Array],
+    outputs: dict[str, Array],
+    drop_unused: bool = False,
+    *,
+    input_prefix: str = "",
+    output_prefix: str = "",
 ) -> onnx.ModelProto:
     """Build and ONNX model from the provided argument-Arrays and outputs.
 
@@ -20,14 +25,20 @@ def build(
         Inputs of the model
     outputs
         Outputs of the model
+    drop_unused
+        Drop unused inputs from the model.
+    input_prefix
+        Prefix to prepend to all input names in the graph.
+    output_prefix
+        Prefix to prepend to all output names in the graph.
 
     Returns
     -------
     onnx.ModelProto
         ONNX model
     """
-    ins = _arrays_to_vars(inputs)
-    outs = _arrays_to_vars(outputs)
+    ins = _arrays_to_vars(inputs, prefix=input_prefix)
+    outs = _arrays_to_vars(outputs, prefix=output_prefix)
 
     mp = spox_build(ins, outs, drop_unused_inputs=drop_unused)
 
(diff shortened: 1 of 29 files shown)


ndonnx/_schema.py
```diff
@@ -52,6 +52,8 @@ class SchemaV1:
     input_schema: dict[str, DTypeInfoV1]
     output_schema: dict[str, DTypeInfoV1]
     version: Literal[1]
+    input_prefix: str | None = None
+    output_prefix: str | None = None
```
cbourjau, line 57: I think there is no harm in simply setting these to `*_prefix: str = ""`. We also don't need to filter them out when serializing. Explicit is great :)

ndonnx/_schema.py
cbourjau, line 51: Before releasing this we need to quickly check if the pydantic parsers in the current `quantcore.onnx_inference` release is ok with the extra fields. Or better yet: we should actually make a release there prior to here.

tests/test_build_utils.py
```diff
@@ -143,3 +143,45 @@ def test_unused_inputs_not_in_schema():
...
+    # Check that it runs successfully
+    import numpy as np
+    import onnxruntime as ort
+
+    session = ort.InferenceSession(mp.SerializeToString())
+    (res,) = session.run(
+        ["out_c"],
+        {"in_a": np.array([1.0, 2.0]), "in_b": np.array([3.0, 4.0])},
+    )
+    np.testing.assert_array_equal(res, np.array([4.0, 6.0]))
```
cbourjau:
```suggestion
```

I think we can leave that part out.

ndonnx/_build.py
```diff
@@ -20,14 +25,20 @@ def build(
...
 
     Returns
     -------
     onnx.ModelProto
         ONNX model
     """
-    ins = _arrays_to_vars(inputs)
-    outs = _arrays_to_vars(outputs)
+    ins = _arrays_to_vars(inputs, prefix=input_prefix)
+    outs = _arrays_to_vars(outputs, prefix=output_prefix)
```
cbourjau: How about we rename in the `{inputs,outputs}` dictionaries directly. Then all that prefix logic is contained to this function and we don't need to touch `_arrays_to_vars`.

cbourjau (dismissed):
Looks great! Thanks! I only have a few minor comments :)

ndonnx/_schema.py
EmilyMohellenbrandQC: The [qc.onnx_inference PR #131 ](https://github.com/Quantco/quantcore.onnx_inference/pull/131) is in draft mode. There the pytest is not working because it expects the ndonnx.build new interface. 

If it is to be released first, how can I make it use the ndonnx branch in CI job?

ndonnx/_schema.py
EmilyMohellenbrandQC: fixed.

tests/test_build_utils.py
EmilyMohellenbrandQC: removed.

ndonnx/_build.py
EmilyMohellenbrandQC: fixed.

ndonnx/_schema.py
EmilyMohellenbrandQC: Updated the qc.onnx_inference PR accordingly. Agree, let's release onnx_inference first.

ndonnx/_build.py
cbourjau: I'm afraid not quite. The prefixing is still happening redundantly. Can't we essentially override the `inputs` and `outputs` variables once in the beginning and be done?

cbourjau (changes requested):

ndonnx/_build.py
EmilyMohellenbrandQC: Screenshot 2026-06-15 at 11 48 30
This is what it looks like since my last commit. The prefixing is done at the beginning of the function and there is no change to the `_arrays_to_vars` function. I don't see `The prefixing is still happening redundantly.`

could you be more specific?

EmilyMohellenbrandQC: I see you requested changes. but I can't see/find it. So I'm wondering are we looking at the same code (latest commit)?

cbourjau: Just below the section shown in the screenshot, there were redundant applications of the "prefixing logic". I fixed it in the way I had in mind in https://github.com/Quantco/ndonnx/pull/218/changes/ddd52da5bd30fe4d4c523ef17ecb88f158c85ffa now.

cbourjau (approved):

ndonnx/_build.py
EmilyMohellenbrandQC: Thanks for fixing it!

PR #223, opened 2026-07-15
Require py311

### LINKED ISSUE #174: Drop python 3.10
After checking a few of our users will be fine with this, we should aim to drop Python 3.10 support. See https://scientific-python.org/specs/spec-0000/.

cbourjau:
Depends on #217 

Closes #174


pixi.toml
```diff
@@ -72,22 +72,23 @@ typos = "*"
...
 numpy = "2.0.*"
 onnxruntime = "==1.20.1"
 
 [environments]
 default = ["test", "lint"]
 default-py311 = ["py311", "test", "lint"]
-py310 = ["py310", "test"]
 py311 = ["py311", "test"]
 py312 = ["py312", "test"]
+py313 = ["py313", "test"]
```
sgilm, line 93: @cbourjau these are not currently used. Did you want to keep them in?

sgilm: Also had to remove `py314` since we still have `osx-64` as a platform. Could that be dropped at this point?

cbourjau: They are nice to have, even if unused, since they ensure that an environment can be solved with these constraints. We should drop `osx-64` though, since its support in upstream `onnx` has been dropped, too.

PR #225, opened 2026-07-30
Add dtype aliases

sgilm:
Adds the following aliases for dtypes (similar to how numpy does it) everywhere there is a `dtype` argument.
- `bool` --> `ndx.bool`
- `int` --> `ndx.int64`
- `float` --> `ndx.float64`
- String versions of the above

This allows for easier duck-typing, e.g.

```python
npx.asarray([1, 2], dtype=bool)  # now works for both npx == numpy (numpy.bool) and npx == ndonnx (ndonnx.bool)
```

### DIFF
--- a/ndonnx/_typed_array/funcs.py
+++ b/ndonnx/_typed_array/funcs.py
@@ -20,9 +20,11 @@ TY_ARRAY_BASE_co = TypeVar("TY_ARRAY_BASE_co", bound="TyArrayBase", covariant=Tr
 
 PYTHON_DTYPE_MAP: dict[PyScalarType, DType] = {
     bool: onnx.bool_,
+    "bool": onnx.bool_,
     int: onnx.int64,
+    "int": onnx.int64,
     float: onnx.float64,
-    str: onnx.utf8,
+    "float": onnx.float64,
 }
 
 
--- a/ndonnx/_typed_array/funcs.py
+++ b/ndonnx/_typed_array/funcs.py
@@ -11,38 +11,12 @@ import numpy as np
 from spox import Var
 
 from ndonnx import DType, from_numpy_dtype
-from ndonnx.types import DTypeAlias, NestedSequence, PyScalar
+from ndonnx.types import NestedSequence, PyScalar
 
 from . import TyArrayBase, datetime, masked_onnx, onnx, promote
 from ._utils import validate_op_result
 
 TY_ARRAY_BASE_co = TypeVar("TY_ARRAY_BASE_co", bound="TyArrayBase", covariant=True)
-
-PYTHON_DTYPE_MAP: dict[DTypeAlias, DType] = {
-    bool: onnx.bool_,
-    "bool": onnx.bool_,
-    int: onnx.int64,
-    "int": onnx.int64,
-    float: onnx.float64,
-    "float": onnx.float64,
-}
-
-
-@overload
-def map_python_dtype(dtype: None) -> None: ...
-
-
-@overload
-def map_python_dtype(dtype: DType | DTypeAlias) -> DType: ...
-
-
-def map_python_dtype(dtype: DType | DTypeAlias | None) -> DType | None:
-    """Intercept a ``dtype`` argument to support Python type aliases."""
-    if dtype is None or isinstance(dtype, DType):
-        return dtype
-    if mapped_dtype := PYTHON_DTYPE_MAP.get(dtype):
-        return mapped_dtype
-    raise TypeError(f"unrecognized dtype argument: `{dtype}`")
 
 
 def _infer_sequence(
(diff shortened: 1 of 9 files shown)


ndonnx/_typed_array/funcs.py
cbourjau, line 1: Can we move this out of the `_typed_array` module? Everything in here should be private. I would hope that we can do all necessary normalization in `_array.py` (and maybe `_funcs.py`), while internal functions can expect the dtype to already be normalized.

ndonnx/types.py
```diff
@@ -26,6 +26,7 @@
 
 PyScalar = bool | int | float | str
 NestedSequence = Sequence["PyScalar | NestedSequence"]
+PyScalarType: TypeAlias = type[bool] | type[int] | type[float] | type[str]
```
cbourjau: I see that it is tempting to reuse the `PyScalar` terminology, but I think we are better off defining something new here. If we introduce dtype aliases, it seems reasonable to support the `Literal` versions too (e.g. `"int64"`), but then `PyScalarType`'s naming would fall apart. 

I suggest going for `DTypeAlias` instead. 

Not having the implied symmetry to `PyScalar` also means we have no pressure to support `str` at this point, which I think we should leave out for now. NumPy's aliases `str` to a fixed-width type:

```
>>> arr = np.asarray([1, 2], dtype=np.uint8).astype(str)
>>> arr
array(['1', '2'], dtype='<U3')
>>> arr[0] = "Hello World!"
>>> arr
array(['Hel', '2'], dtype='<U3')
```

cbourjau (changes requested):
Awesome! I was about to tackle this, too. Thanks for taking the lead! I think we should support the string aliases (e.g., `"int64"`), too, but leave out `str` for now. This will be a very nice usability improvement!

ndonnx/_typed_array/funcs.py
sgilm: I moved it into `_dtypes.py`, but maybe it's best not to mix the `DType` definition with references to concrete `DType`s. It needed a lazy import to work. Did you have a better place in mind?

ndonnx/types.py
sgilm: Makes perfect sense!

Are you proposing the whole suite of string aliases? E.g. all "int64", "int32", "float64", "float32" etc.?

What about the numpy objects, e.g. `np.int64`? I guess if we are using the string aliases and we have one common way of specifying equivalent dtypes across `numpy/ndonnx`, adding `np.int64` etc. too is not that helpful.

ndonnx/types.py
```diff
@@ -27,6 +27,15 @@
 PyScalar = bool | int | float | str
 NestedSequence = Sequence["PyScalar | NestedSequence"]
 
+DTypeAlias: TypeAlias = (
+    type[bool]
+    | type[int]
+    | type[float]
+    | Literal["bool"]
+    | Literal["int"]
+    | Literal["float"]
```
cbourjau:
```suggestion
    | Literal["bool", "int", "float"]
    | Literal["int8", "int16", "int32", "int64"]
    | Literal["uint8", "uint16", "uint32", "uint64"]
    | Literal["float16", "float16", "float16"]
```
I'd be in favor of doing all the default, non-controversial aliases in this PR. WDYT?

ndonnx/_typed_array/funcs.py
cbourjau: How about putting `normalize_dtype` into `ndonnx/_funcs.py`? That module should be at the bottom of the dependency graph, where `import ndonnx as ndx` is fully available.

tests/test_creation.py
```diff
@@ -94,3 +94,35 @@ def do(npx):
...
+    [
+        (bool, ndx.bool),
+        ("bool", ndx.bool),
+        (int, ndx.int64),
+        ("int", ndx.int64),
+        (float, ndx.float64),
+        ("float", ndx.float64),
+    ],
+)
+def test_dtype_aliases(dtype_alias, ndx_dtype):
```
cbourjau: Wrt testing this: Mypy should have us covered pretty well here - if we annotate our public interface correctly. For consistency, we should allow the alias to be used in all public functions in `ndonnx`, `ndonnx.Array`, and `ndonnx.extensions`. We could add a unit test that iterates through the interfaces, looks for an argument called `dtype,` and if found, checks that we annotate it using `ndonnx.DType | DTypeAlias`.

cbourjau (changes requested):
Sorry, a few more minor comments 🙈. Could you also add a changelog entry, please?

ndonnx/types.py
sgilm: Sounds great to me!

ndonnx/types.py
cbourjau: Let's not allow `np.int64` for now because I don't think we can type them as safely as we can the string aliases. Something like `Literal[np.int64]` is not valid.

ndonnx/_typed_array/funcs.py
sgilm: Looks like this creates a circular dependency between `ndonnx/_funcs` and `ndonnx/_array`, but we get away with a single lazy import in `_array`, so not too bad.

tests/test_creation.py
sgilm: Great, I added a `test_namespace` file for this and covered `ndonnx` and `ndonnx.Array`. On `ndonnx.extensions`: most of the functions accepting `dtype` are e.g. `is_nullable_dtype` etc.

Not sure if it makes sense to allow the aliases there, since they don't map so well to `numpy`. But we could also just be consistent with aliases regardless, and apply them everywhere.

cbourjau: Ok, let's leave `ndonnx.extensions` out. It is not duck-typable with NumPy, and thus, in order to use it, users (a) know that they are dealing with ndonnx arrays and (b) have the ndonnx namespace available.

cbourjau (approved):
Beautiful! Thanks again!

PR #231, opened 2026-09-05
Support NumPy scalars across public APIs

### LINKED ISSUE #224: Add typing support for `numpy.generic` scalars
In various parts of the codebase, `PyScalar = bool | int | float | str` is used as the type hint for scalar values. Using a `numpy.generic` in place of a `PyScalar` mostly works at runtime, but doesn't satisfy type checkers.

```python
import numpy as np
import ndonnx as ndx

# Passes type check.
ndx.asarray(3)  # array(data: 3, dtype=int64)

# Doesn't pass type check.
ndx.asarray(np.int64(3))  # array(data: 3, dtype=int64)
```

This should hopefully be easy to support on a sensible subset of `numpy.generic` scalar types.

Danila-Pechenev:
Closes #224.

## Summary

This PR expands support for NumPy scalar values across the relevant public scalar-data and operand APIs.

- Adds explicit scalar type aliases for the supported NumPy primitive scalar families while preserving the existing Python scalar types.
- Treats supported NumPy scalars as strong, dtype-bearing operands while preserving existing weak-scalar semantics for Python scalars.
- Extends NumPy scalar support across creation, promotion, selection, elementwise operations, Array operators, manipulation/reduction parameters, temporal construction/ranges, and extensions.
- Preserves existing ndarray operands, custom DType dispatch, and `np.datetime64` / `np.timedelta64` operator behavior.
- Adds backend compatibility where required for float16 LabelEncoder inputs and int16 mapping outputs.
- Adds runtime and static-typing regression coverage for the supported scalar categories.

Unsupported NumPy scalar categories such as complex and void remain outside the supported scalar set.

## Testing

Added regression coverage for, among other cases:

- NumPy scalar construction and nested scalar values.
- Strong NumPy versus weak Python scalar promotion.
- `result_type` ordering and equivalent rank-zero operands.
- Arithmetic, comparison, bitwise, logical, and selection APIs.
- NumPy boolean arithmetic operands.
- Integer and floating `arange` edge cases and overflow handling.
- Temporal construction, assignment, and inferred `arange`.
- NumPy integer `repeat` values.
- ndarray operand backward compatibility.
- `isin` NaN handling and float16 inputs.
- `static_map` scalar keys and backend-supported outputs.
- Custom DType construction, promotion, and range hooks.


cbourjau:
#224 is not only about making asarray work. If we add support, we want to apply it to all public functions to ensure consistency. It's not very complicated, but it's not entirely trivial either.

Danila-Pechenev:
@cbourjau  Thanks for the clarification! I’ve expanded the PR accordingly.

NumPy scalar support is now applied consistently across the relevant public scalar-data and operand APIs, rather than only to asarray. NumPy scalars participate as strong, dtype-bearing operands while the existing weak-scalar semantics for Python scalars are preserved.

I’ve also added runtime and mypy regression coverage for the affected API categories, including promotion, elementwise operations, Array operators, temporal ranges, extensions, ndarray compatibility, and custom DType hooks.

Could you take another look when you have a chance?

cbourjau:
Thanks for taking a stab at this. I think I found a way to do this with a smaller diff in #241, which should supersed this workl

PR #234, opened 2026-09-08
Add numerically sound implementations for log1p and expm1

cbourjau:
We recently ran into the lack of these functions when working on some internal documentation. The implementation is tested by the upstream array-api test suite. It turns out that onnxruntime's `Where`-operator apparently doesn't correctly propagate `-0.0`, which is why those special cases are still skipped.

FYI @fjetter

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@
 Changelog
 =========
 
+0.22.0 (unreleased)
+-------------------
+
+**New feature**
+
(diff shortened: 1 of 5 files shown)


fjetter (approved):
thank you!

cbourjau:
Upstream issue for the mentioned -0.0 issue with `Where`: https://github.com/microsoft/onnxruntime/issues/32191

PR #235, opened 2026-09-08
Build models expecting the operator support of onnxruntime >=1.26.0

cbourjau:
We deploy many workarounds for bugs and missing kernel support in onnxruntime. Currently, we build models targeting onnxruntime >= 1.20.1. We couldn't update this for a long time due to various regressions upstream. However, onnxruntime >=1.26.0 is finally in a good enough shape that we can rely on it. Note that upstream is already on 1.29.2, but there are no fixes in those newer versions relevant to these workarounds, so we can target 1.26.0 without missing anything for now.

The possibly most impactful removed workaround is for a bug in basic arithmetic operations on zero-sized arrays, which could lead to segfaults in onnxruntime. We mitigated that using an unsqueeze-op-squeeze pattern, which significantly increased graph size and even runtime in some circumstances.

Note that this long-overdue update is not an indication of how conservative we want to be about this in the future. Ideally, we will update the minimally expected onnxruntime version within about a month after a new relevant onnxruntime version becomes available on conda-forge.

### DIFF
--- a/ndonnx/_typed_array/ort_compat.py
+++ b/ndonnx/_typed_array/ort_compat.py
@@ -2,12 +2,12 @@
 # SPDX-License-Identifier: BSD-3-Clause
 """Compatibility layer to work around missing kernels in onnxruntime.
 
-Currently targets 1.20.1:
-https://github.com/microsoft/onnxruntime/blob/v1.20.1/docs/OperatorKernels.md
+Currently targets 1.26.0:
+https://github.com/microsoft/onnxruntime/blob/v1.26.0/docs/OperatorKernels.md
 
 Updates to this file may be informed by inspecting the diff for ``OperatorKernels.md``
 between two tags (e.g.
-https://github.com/microsoft/onnxruntime/compare/v1.19.0..v1.20.1/).
+https://github.com/microsoft/onnxruntime/compare/v1.20.1..v1.26.0/).
 """
 
 from __future__ import annotations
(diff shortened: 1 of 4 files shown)


ndonnx/_typed_array/ort_compat.py
```diff
@@ -488,8 +460,13 @@ def top_k(
     largest: int = 1,
     sorted: int = 1,
 ) -> tuple[Var, Var]:
-    # tensor(double), tensor(float), tensor(int32), tensor(int64)
-    mapping = _common_mapping
+    # tensor(double), tensor(float), tensor(float16), tensor(int16),
```
sgilm: It looks like `float16` [may not be](https://github.com/microsoft/onnxruntime/blob/8c546c37b43caaca1fa25db430dab94b901cf277/docs/OperatorKernels.md?plain=1#L508) supported?

ndonnx/_typed_array/ort_compat.py
sgilm: From the comment and `onnxruntime` PR, should this check just be for an `int64` (like it is in `min`)?

sgilm (approved):

ndonnx/_typed_array/ort_compat.py
cbourjau: Good catch! Turns out onnxruntime has a special pass that inserts float16-to-float32 casts when there is no kernel. I added a comment to that for future reference.

PR #238, opened 2026-09-16
Support "datetime64[{s,ms,us,ns}]" and "timedelta64[{s,ms,us,ns}" aliases

### LINKED ISSUE #236: Provide dtype aliases for string and datetime types
The `argument` and `normalize_dtype` functions can map the several float, int and bool types from string representations (`"int16" -> onnx.int16`), but these functions lack aliases for the string and `ndo.DateTime64DType` types.

For example, to create a datetime argument, I have to use `ndo.argument(shape=("N", 1), dtype=ndo.DateTime64DType(unit="us"))`. I would find it easier to just call `ndo.argument(shape=("N", 1), dtype="datetime64[us]")`  instead.

Same problem when creating string arguments. I would like to just call `ndo.argument(shape=("N", 1), dtype="string")`.

Where does this come up? When creating arguments from the metadata of variable classes.

cbourjau:
Add string aliases for all datetime and timedelta dtypes currently supported by ndonnx. This closes #236 .


lutzgruber-quantco (approved):

CHANGELOG.rst
```diff
@@ -7,12 +7,13 @@
 Changelog
 =========
 
-0.22.0 (unreleased)
+0.22.0 (2026-09-16)
```
sgilm:
```suggestion
0.22.0 (2026-09-17)
```

sgilm (approved):
It now occurs to me that we also have `from_numpy_dtype` doing a similar job to `DTYPE_ALIAS_MAP` and `normalize_dtype`. Do you think it would be worth cleaning that up in a future PR? If so I'd be happy to do that.

cbourjau:
> It now occurs to me that we also have from_numpy_dtype doing a similar job to DTYPE_ALIAS_MAP and normalize_dtype. Do you think it would be worth cleaning that up in a future PR? If so I'd be happy to do that.

Yes, please ❤️ !

PR #239, opened 2026-09-18
Support `isin` for time-like data types

sgilm:
Adds support for `ndonnx.extensions.isin` applied to arrays with datetime or timedelta data types.

I only allowed comparisons between `ndx.Datetime64Dtype`/`np.datetime64` and `ndx.TimeDelta64Dtype`/`np.timedelta64` here. Numpy may also produce positive matches between `np.timedelta64` and regular `ints` but this behaviour is deprecated. Any other comparisons in Numpy produce `False`, and also has the possibility for some quite bizarre bugs:

```python
values = np.array([1, 2], dtype="datetime64[ns]")

# [False, False]: int and datetime are not compatible.
np.isin(values, [1])

# [True, False]: int now matches, but datetime doesn't...
np.isin(values, [1, np.datetime64(2, "ns")])

```


ndonnx/_typed_array/onnx.py
```diff
@@ -958,7 +956,12 @@ def __ndx_equal__(self, other) -> TyArrayBool:
             return TyArrayBool(var)
         return NotImplemented
 
-    def isin(self, items: Sequence[VALUE]) -> TyArrayBool:
+    def isin(self, items: Sequence[ISIN_SCALAR]) -> TyArrayBool:
+        if not is_non_time_seq(items):
+            raise ValueError(
+                f"unexpected type in 'items' for 'isin' on `{self.dtype}`: `{items}`"
+            )
```
sgilm: It's tempting to think this could be generalized for other incompatible types in the hierarchy (e.g. 'utf8' and numeric) so we would get a similar error message rather failing on an attempt to cast, but I can see how that's a lot of work for no real benefit.

sgilm (commented):
Thanks for fixing up the typing :)

ndonnx/_typed_array/onnx.py
cbourjau: I'd love to hear more if you have an idea for this!

cbourjau (approved):

